PR #4, opened 2020-05-02
Upgrade async code

vorot93:
Replaces poll_fn and futures::select+timeout combo with `tokio::select!`.


AgeManning:
Awesome. Thanks. Over time I'll get to upgrading the main code

PR #12, opened 2020-05-20
Expose underlying libraries for external use

AgeManning:
Some extension projects require access to some underlying modules. This PR exposes these libraries.


AgeManning:
Already resolved.

PR #30, opened 2020-08-16
Add openssl-vendored feature again

paulhauner:
This seems to have been lost in the latest version and it's preventing Lighthouse from compiling.


paulhauner:
Sorry, didn't realise openssl was removed.

PR #33, opened 2020-08-27
cache hash(local_node_id)

### LINKED ISSUE #32: hash(local_node_id) can be cached
Each time a `RandomPacket` is received, the src_node_id must be calculated from the tag field.  This requires the hash(local_node_id) to be calculated:

https://github.com/sigp/discv5/blob/71fa57295cc1c31b3f5361f6d06e3f9ab0b9b06a/src/handler/mod.rs#L932

This can be avoided if the hash(local_node_id) is cached.

jrhea:
added logic for caching hash(local-node-id) and test for src_id method. this was done to avoid hash calc that happens when a client is repeatedly sent `RandomPackets` 👀

resolves #32

@AgeManning lmk if i need to change anything or if you want more context

### DIFF
--- a/src/handler/mod.rs
+++ b/src/handler/mod.rs
@@ -33,6 +33,7 @@ use futures::prelude::*;
 use log::{debug, error, trace, warn};
 use lru_time_cache::LruCache;
 use parking_lot::RwLock;
+use sha2::digest::generic_array::{typenum::U32, GenericArray};
 use sha2::{Digest, Sha256};
 use std::sync::Arc;
 use std::{collections::HashMap, default::Default, net::SocketAddr, sync::atomic::Ordering};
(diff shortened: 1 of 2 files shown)


AgeManning (approved):
Oh yeah cool. 

I wonder if you did any benchmarks for this? 

In either case, if we can avoid the hash per packet, its a win :). 

Note, that there is a filter (in the recv side of the socket) that limits incoming packets so we shouldn't able to be flooded with packets at this layer.

jrhea:
> Note, that there is a filter (in the recv side of the socket) that limits incoming packets so we shouldn't able to be flooded with packets at this layer.

@AgeManning ya I definitely noticed a difference in this implementation vs others based on this.  I've been meaning to get your thoughts on rate limiting.  I'll ping u on tg.

PR #38, opened 2020-09-17
Remove unnecessary verification

ackintosh:
In the function [handle_challenge](https://github.com/sigp/discv5/blob/839254cfb8fc3cd6bb3176393839793e31f0303d/src/handler/mod.rs#L512), both of the variables (enr, node_address) points to the same instance of NodeContact:
- enr comes from [the request_call.contact](https://github.com/sigp/discv5/blob/839254cfb8fc3cd6bb3176393839793e31f0303d/src/handler/mod.rs#L601)
- node_address comes from [the request_call.contact](https://github.com/sigp/discv5/blob/839254cfb8fc3cd6bb3176393839793e31f0303d/src/handler/mod.rs#L543-L546)

So I think that [the verification](https://github.com/sigp/discv5/blob/839254cfb8fc3cd6bb3176393839793e31f0303d/src/handler/mod.rs#L604) can be omitted. 💡

### DIFF
--- a/src/handler/mod.rs
+++ b/src/handler/mod.rs
@@ -600,39 +600,27 @@ impl Handler {
         // Check if we know the ENR, if not request it and flag the session as awaiting an ENR.
         match request_call.contact.clone() {
             NodeContact::Enr(enr) => {
-                // Verify the ENR and establish or fail a session.
-                if self.verify_enr(&enr, &node_address) {
-                    // Send the Auth response
-                    trace!(
-                        "Sending Authentication response to node: {}",
-                        request_call
-                            .contact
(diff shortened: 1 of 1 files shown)


AgeManning:
Yes, this looks right. The verification is not necessary as we're using the outbound identification. 

I think this is an issue with the spec, in that we can't id the source in challenges. 

I'm currently writing the update to v5.1. In v5.1 this check will be valid. 

If you can fix the fmt formatting issues and get tests to pass, will merge this in. 

Thanks!

ackintosh:
Thanks for your review! I've fixed the formatting issue.

PR #39, opened 2020-09-22
Upgrade to discovery v5.1

AgeManning:
# Description

This upgrades to the v5.1 (as it currently stands in the spec repo). 

This shifts these repo to the beta semver channel. v5.0 will still be temporarily maintained in the alpha versioning. 

This currently still requires interop testing.


fjl:
Unfortunately, I can't build this because there is a syntax error.

AgeManning:
This should now build and should be up to date with the 5.1 spec :crossed_fingers:

AgeManning:
There seems to still be some discrepencies in RLP encoding. And the talkreq and talkresp methods still need to be added.

PR #48, opened 2020-10-30
Drop a packet with invalid packet size

### LINKED ISSUE #47: The limit of packet size
From [the wire protocol](https://github.com/ethereum/devp2p/blob/master/discv5/discv5-wire.md#udp-communication), there's a limited range of packet size which implementations should consider the packet is valid:

- the maximum size of any packet is 1280 bytes
- the minimum size of any packet is 63 bytes

Currently we support the limit [partially](https://github.com/sigp/discv5/blob/f117b3ca56fa3dca2317270434634ff7106d391a/src/socket/recv.rs#L17), not sufficient. I'll file a PR to improve that. 😃

ackintosh:
This PR closes #47

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -50,6 +50,8 @@ impl From<std::io::Error> for Discv5Error {
 pub enum PacketError {
     /// The packet type is unknown.
     UnknownPacket,
+    /// The packet size was larger than expected.
+    TooLarge,
     /// The packet size was smaller than expected.
     TooSmall,
     /// The NodeId sent was invalid.
--- a/src/error.rs
+++ b/src/error.rs
@@ -45,7 +45,7 @@ impl From<std::io::Error> for Discv5Error {
     }
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
 /// Types of packet errors.
 pub enum PacketError {
     /// The packet type is unknown.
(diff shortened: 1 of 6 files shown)


AgeManning (commented):
Thanks for looking closely at the code. 

Both the cases listed in #47 are already handled. The buffer size is 1280, limiting the maximum bytes we can accept. 

In terms of checking outbound traffic, the maximum is set because the application layer in principle could send too many ENR's in some RPC so we double check on the outbound if it exceeds the maximum byte length. 

The reason we don't check the minimum is because the minimum is based on the basic encryption wire format. Regardless of the application layer, it should be impossible to send less than the minimum unless the encoding is broken and not to spec. For this reason, checking the minimum on the outbound was explicitly left out. 

If you think we should make the max check more specific by making it a const variable as you've done, I'm happy to add that in. 

Let me know what you think

ackintosh:
Thank you for your comment! Now I understand that the cases are already handled, and why checking the minimum of outbound traffic was left out.

Yes, I believe it's worth to add the max check. It improves simplicity of the software.

Is it better to file an another PR to add the max check? Since this PR has unnecessary changes, it makes noise to add commits to remove the unncessary changes.

AgeManning:
I'm happy to do what you prefer. 

Feel free to make changes to this PR or close it and make another. 

We will squash merge to master regardless, so extra commits will be removed from history,

ackintosh:
I've add commits to remove extra changes to this PR. Please have another look. 😃

AgeManning (approved):
Nice thanks

PR #63, opened 2021-02-21
Auto-coerce to ping response to ipv4 when possible

paulhauner:
If a peer sends a ping response with an ipv6 address that is ipv4 compatible/mapped, then auto-convert to ipv4.

Related to https://github.com/sigp/discv5/pull/62


AgeManning (approved):
LGTM

PR #77, opened 2021-07-20
Fix issues in `update_value`

michaelsproul:
This PR fixes an issue in `KBucket::update_value` that caused two nodes to be removed rather than one when the filter check failed. It also updates the value of `first_connected_pos` so that it remains consistent in the case of a removal.

Testing has been expanded through a new `quickcheck` test which updates a value in the presence of a filter. I've also re-jigged some of the other quickcheck infra to make use of an `Arbitrary`-aware generator for `NodeId`, and refactored some of the common checks into `KBucket::check_invariants`.


AgeManning (approved):
Nice! Good find!

PR #78, opened 2021-07-21
Add getter function for node_address

KolbyML:
This just adds a getter for node_address: NodeAddress in Impl TalkRequest.

This is useful if you want to send a talk request to a node that sent you one. Because you can find their ENR from it.

### DIFF
--- a/src/service.rs
+++ b/src/service.rs
@@ -80,6 +80,10 @@ impl Drop for TalkRequest {
 }
 
 impl TalkRequest {
+    pub fn node_id(&self) -> &NodeId {
+        &self.node_address.node_id
+    }
+
     pub fn protocol(&self) -> &[u8] {
         &self.protocol
     }


AgeManning:
Thanks for the addition. 

The `NodeAddress` type is something I made which was meant to be just used internally. I think this object may not be exposed to users at the moment. Do you think the socket address is important also? 

We could either:
1. Make `NodeAddress` a public object and return it (as you've done here)
2. Just use the `NodeId`, which should also be sufficient for searching for an ENR. 

I'm not entirely sure about your use-case, do you think 2 is reasonable?

KolbyML:
> Thanks for the addition.
> 
> The `NodeAddress` type is something I made which was meant to be just used internally. I think this object may not be exposed to users at the moment. Do you think the socket address is important also?
> 
> We could either:
> 
> 1. Make `NodeAddress` a public object and return it (as you've done here)
> 2. Just use the `NodeId`, which should also be sufficient for searching for an ENR.
> 
> I'm not entirely sure about your use-case, do you think 2 is reasonable?

@AgeManning Yeah I totally think that is reasonable saying that at the moment I only want the NodeId anyways to search for an ENR as you said, I don't really need the socket address.

I updated my PR to reflect the changes seen in option two.

AgeManning (approved):
Looks good to me.

AgeManning:
@KolbyML Just wanted to point out some things. 

The routing table won't always contain the ENR of a node that has performed this request. Although we do request ENR's and verify their validity in the handshake, there are circumstances where we can receive a request without knowing the ENR of the sender. 

More commonly however, is that we only add the ENR to the local routing table if is not full. On a large network, we'd end up dropping most of the ENRs we come across. So if you are searching the routing table, it might a low probability of finding that ENR. 

Potentially we could add an event that emits the ENR when we establish connections and the user could store them externally in a data structure that's easier to search and the user can handle the size limits etc of that.

KolbyML:
> @KolbyML Just wanted to point out some things.
> 
> The routing table won't always contain the ENR of a node that has performed this request. Although we do request ENR's and verify their validity in the handshake, there are circumstances where we can receive a request without knowing the ENR of the sender.
> 
> More commonly however, is that we only add the ENR to the local routing table if is not full. On a large network, we'd end up dropping most of the ENRs we come across. So if you are searching the routing table, it might a low probability of finding that ENR.
> 
> Potentially we could add an event that emits the ENR when we establish connections and the user could store them externally in a data structure that's easier to search and the user can handle the size limits etc of that.

@AgeManning Thank you for the insight. Currently I am working to implement uTP as a sub-protocol for TalkRequests so once I get to it I will probably look into implementing what you went over thank you for the help :)

PR #82, opened 2021-08-04
Move kbucket into separate DHT crate

jacobkaufmann:
- Create workspace with members `discv5` and `discv5-dht`
- Move kbucket into `discv5-dht` package
- Move ConnectionDirection from handler module to kbucket module

In order to build overlay networks on top of discv5, we need to be able to build and manage separate routing tables. To support this, we need the kbucket items to be public.


ogenev:
What are the benefits of creating a separate `discv5-dht` package? If we want to be able to re-use the kbucket module, I think we can just change `mod kbucket`; to `pub mod kbucket` in `discv5/src/lib.rs`?

jacobkaufmann:
> What are the benefits of creating a separate `discv5-dht` package? If we want to be able to re-use the kbucket module, I think we can accomplish this by changing `mod kbucket;` to `pub mod kbucket;` in `discv5/src/lib.rs`?

The benefit I see is that it may not make sense to make `kbucket` a public module within the `discv5` package. The public interface for `discv5` does not necessarily need to expose the underlying Kademlia DHT items. The separation is mainly logical, because I think it makes more sense for those items to be made public from their own crate.

AgeManning:
I'm not too concerned with making the DHT items public and think it will be a simpler modification.

If we are worried about the public API, we could still gate the public access via a feature flag. Would this be a sufficient solution for your needs?

jacobkaufmann:
Something else that we have in mind is a `Metric` trait to support alternative (non-XOR) distance metrics. We would like to be able to associate a `Metric` with `Key` so that we can define a metric space for the DHT key space.

Given that Kademlia is strongly associated with XOR, I thought it would be better organizationally to separate it out into a separate and more general DHT crate. However, if you don't see an issue with this, then I agree that it will be simpler to keep everything in a single crate.

AgeManning:
Yeah I can see the appeal in adding a `Metric` trait or even a `DHT` trait to keep all the encryption and protocol logic but swap out the underlying DHT. 

The reason these are not configurable at the moment, is because I had planned on keeping this crate to match the discv5 specifications which still require the XOR metric (https://github.com/ethereum/devp2p/blob/master/discv5/discv5-theory.md) (although I do realise you could use any other and still participate on a network). 

If you just need to change the metric and leave other elements of the DHT alone, i'd be happy include a `Metric` trait PR to make it more general. 

If it works for you then, could you modify this PR to keep the single crate and just make the modules you need public. If we need to do anything more extensive we can go from there. Sound good?

jacobkaufmann:
I reverted the initial commit and made the `kbucket` module public. As I get further into the implementation, I may need to expose additional modules, but I did not identify anything else at this time.

And I think there may be a misunderstanding around the motivation for the `Metric` trait. We do not intend to use a different (non-XOR) metric for the discv5 routing table. Alternative metrics would be for application-level overlay network routing tables.

AgeManning:
Ok right. I think i may have misunderstood the reason for the `Metric` trait. 

Anyway, keep making PRs as you go, hopefully we can make this repo as inclusive for your needs as we can. :)

AgeManning (approved):

PR #83, opened 2021-08-11
Emit warn log when no known_closest_peers found

ackintosh:
Added warn log to notice users that no known_closest_peers found. And I think it's better to return empty result immediately in the case.


AgeManning (commented):
Yep I agree. 

Thanks for the addition.

Issue #71, opened 2021-06-15
Re-implement lru-time-cache


AgeManning:
The current `lru_time_cache` dependency is in efficient under most operations.

We are using this cache for a variety of fields in the handler. The sessions in particular can be quite large and we don't want to be doing O(n) operations on every get for every message.

See eg, https://github.com/maidsafe/lru_time_cache/issues/143

ackintosh:
I'm working on this issue, will file a PR in a few days. 😃

PR #84, opened 2021-08-15
Re-implement lru-time-cache

ackintosh:
close https://github.com/sigp/discv5/issues/71

- use `hashlink::LinkedHashMap` as it's better to avoid to write unsafe code
- I have measured performance of LruTimeCache with the benchmark code below. LruTimeCache was `1,142 ns/iter`, while std::collections::HashMap was `664 ns/iter`.
  - It is good score
  - for your reference, I run [this benchmark]( https://github.com/maidsafe/lru_time_cache/issues/143#issue-756136904), lru_cache_sum was `101,426 ns/iter`

> test lru_time_cache::bench::hashmap_sum        ... bench:         664 ns/iter (+/- 195)
> test lru_time_cache::bench::lru_time_cache_sum ... bench:       1,142 ns/iter (+/- 187)

```rust

mod bench {
    extern crate test;

    use test::Bencher;

    use std::cell::RefCell;
    use std::collections::HashMap;

    const RANGE: usize = 32 * 1024;
    const FIND_TIMES: usize = 10;

    use rand::prelude::*;
    use crate::lru_time_cache::LruTimeCache;
    use std::time::Duration;

    #[bench]
    fn lru_time_cache_sum(b: &mut Bencher) {
        let mut lru = LruTimeCache::new(Duration::from_secs(10), Some(RANGE));
        for i in 0..RANGE {
            lru.insert(i, i);
        }
        let lru = RefCell::new(lru);
        b.iter(|| {
            let res: usize = (0..FIND_TIMES)
                .map(|_| {
                    *lru.borrow_mut()
                        .get(&thread_rng().gen_range(0..RANGE))
                        .unwrap_or(&0)
                })
                .sum();
            test::black_box(res);
        });
    }

    #[bench]
    fn hashmap_sum(b: &mut Bencher) {
        let mut map = HashMap::new();
        for i in 0..RANGE {
            map.insert(i, i);
        }
        b.iter(|| {
            let res: usize = (0..FIND_TIMES)
                .map(|_| *map.get(&thread_rng().gen_range(0..RANGE)).unwrap_or(&0))
                .sum();
            test::black_box(res);
        });
    }
}
```

(Could not add the benchmark code to this repo as benchmarking support is unstable.)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,7 @@ aes-gcm = "0.9.2"
 tracing = { version = "0.1.26", features = ["log"] }
 tracing-subscriber = "0.2.19"
 lru = "0.6.5"
+hashlink = "0.7.0"
 
 [dev-dependencies]
 rand_07 = { package = "rand", version = "0.7" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,6 @@ digest = "0.9.0"
 rand = { version = "0.8.4", package = "rand" }
 smallvec = "1.6.1"
 parking_lot = "0.11.1"
-lru_time_cache = "0.11.11"
 lazy_static = "1.4.0"
 aes = { version = "0.7.4", features = ["ctr"] }
 aes-gcm = "0.9.2"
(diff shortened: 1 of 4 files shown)


AgeManning (approved):
This looks great. Thanks!

I think we'll merge this in, but will do some substantial testing before cutting a new release. 

Cheers!

PR #85, opened 2021-08-15
Remove mut self from find_enr()

KolbyML:
- Removed mut self from find_enr()

So I am making this PR because to my knowledge the mut self isn't required and it will work without mutability. I am not the most knowledgeable in Rust, but it does seem to be unnecessary correct me if I am wrong.

It would be nice to have this not mutable because then you can use it from an Arc reference of the Discv5 Impl.

@AgeManning

### DIFF
--- a/src/discv5.rs
+++ b/src/discv5.rs
@@ -284,7 +284,7 @@ impl Discv5 {
     }
 
     /// Returns an ENR if one is known for the given NodeId.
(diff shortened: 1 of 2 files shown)


AgeManning (approved):
Yep this looks correct to me. I thought we had removed these in an earlier PR, looks like we missed some. 

Thanks

KolbyML:
Np, ty for reviewing it quick

PR #89, opened 2021-10-11
Prevent adding duplicate nodes

AgeManning:
With filters enabled, it was possible (although unlikely) to add duplicate nodes to a bucket. 

This can happy via the following set of actions:

- The bucket becomes full
- A node (call it A) gets added to the pending state of the bucket
- A node gets removed from the bucket for updating state that is incompatible with a filter, making the bucket no longer full
- The node A can then get inserted into the bucket from a new RPC message (as the table is no longer full). This happens before the pending time. 
- The pending timeout expires and the pending node (A) gets inserted in the table as a duplicate

This PR drops the pending node if it already exists in the bucket.

### DIFF
--- a/src/kbucket/bucket.rs
+++ b/src/kbucket/bucket.rs
@@ -601,7 +601,6 @@ where
         if matches!(insert_result, InsertResult::Inserted) && inserting_pending {
             self.pending = None
         }
-        }
         insert_result
     }
 


src/kbucket/bucket.rs
```diff
@@ -1127,6 +1146,69 @@ pub mod tests {
...
+        assert!(bucket.pending().is_some());
+
+        // A misc node gets dropped, because it may not pass a filter when updating its connection
+        // status.
+        bucket.remove(&third.key);
+
+        // The pending nodes status gets updated
+        // Apply pending gets called within kbuckets, so we mimic here.
+        // The pending time hasn't elapsed so nothing should occur.
+        println!("Result of apply pending: {:?}", bucket.apply_pending());
```
michaelsproul, line 1190: Could we replace all these `println`s with asserts? It seems the values stay the same between runs:

```
Result of updating first node state: Updated
Result of apply pending: None
Result of inserting node: Inserted
Result of apply pending: None
Result of updating the pending nodes status: NotModified
```

src/kbucket/bucket.rs
```diff
@@ -1127,6 +1146,69 @@ pub mod tests {
         assert_eq!(MAX_NODES_PER_BUCKET - 1, bucket.num_disconnected());
     }
 
+    /// No duplicate nodes can be inserted via the apply_pending function.
+    #[test]
+    fn full_bucket_applied_no_duplicates() {
+        // First fill the bucket with disconnected nodes.
```
michaelsproul, line 1152:
```suggestion
        // First fill the bucket with connected nodes.
```

src/kbucket/bucket.rs
```diff
@@ -1127,6 +1146,69 @@ pub mod tests {
...
+        // status.
+        bucket.remove(&third.key);
+
+        // The pending nodes status gets updated
+        // Apply pending gets called within kbuckets, so we mimic here.
+        // The pending time hasn't elapsed so nothing should occur.
+        println!("Result of apply pending: {:?}", bucket.apply_pending());
+        println!(
+            "Result of inserting node: {:?}",
+            bucket.insert(node.clone())
```
michaelsproul, line 1193: After the insert we should assert that `bucket.pending.is_none()`

src/kbucket/bucket.rs
```diff
@@ -566,6 +568,15 @@ where
...
                 }
+
+                // If we inserted the node, make sure there is no pending node of the same key. This can
+                // happen when a pending node is inserted, a node gets removed from the bucket, freeing up
+                // space and then re-inserted here.
+                if let Some(pending) = self.pending.as_ref() {
+                    if pending.node.key == node.key {
+                        self.pending = None
+                    }
+                }
```
divagant-martian, line 579: This code is the same as in the disconnected case, so if I understand correctly the logic is: if inserted the pending, remove it? If so what would you say of factoring this out as 
`let insert_result =  match node.status.state { .. };
and later checking if it is Inserted. I might be clearer.

divagant-martian: we should update the docs to say that if the inserted node matches the pending one, this one is dropped.

src/kbucket/bucket.rs
michaelsproul: ^I think this should replace the `if let Some()` speed-up block

src/kbucket/bucket.rs
```diff
@@ -545,7 +547,9 @@ where
             }
         }
 
-        match node.status.state {
+        let key = node.key.clone();
```
divagant-martian, line 550:
```suggestion
        let inserting_pending = self
            .pending
            .as_ref()
            .map(|pending| pending.node.key == node.key)
            .unwrap_or_default();
```

src/kbucket/bucket.rs
```diff
@@ -583,14 +589,27 @@ where
...
+
+        // If we inserted the node, make sure there is no pending node of the same key. This can
+        // happen when a pending node is inserted, a node gets removed from the bucket, freeing up
+        // space and then re-inserted here.
+        if matches!(insert_result, InsertResult::Inserted) {
+            if let Some(pending) = self.pending.as_ref() {
+                if pending.node.key == key {
+                    self.pending = None
+                }
+            }
```
divagant-martian, line 602:
```suggestion
        if matches!(insert_result, InsertResult::Inserted) && inserting_pending {
            self.pending = None
        }
```

PR #92, opened 2021-11-09
Allowing multithreaded Discv5 calls, without locking

carver:
We're using `Discv5` in trin, in a multithreaded (async) context. I find it somewhat painful to have to lock the `Discv5` instance to use most of the APIs.

My read is that, internally, `Discv5` already locks the internal values to make for safe multi-threaded usage. I think we can make that apparent/accessible in the API by just converting the `&mut self` to `&self` everywhere possible (basically everywhere except for startup and shutdown, both of which are not a problem for us).

I'm curious what you folks think about this approach, or if you have suggestions for how trin could use a `Discv5` instance across threads without locking it at every invocation.

### DIFF
--- a/src/discv5.rs
+++ b/src/discv5.rs
@@ -174,7 +174,7 @@ impl Discv5 {
     /// addresses, so that they can be used immediately in following DHT
     /// operations involving one of these peers, without having to dial
     /// them upfront.
-    pub fn add_enr(&mut self, enr: Enr) -> Result<(), &'static str> {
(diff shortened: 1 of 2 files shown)


AgeManning:
Yes you're right. In its original version I had most of the calls semi-schronous and switched later to async, but obviously have missed the unnecessary &mut self. 

The suggested changes look fine to me

carver:
Great! Let me know if there's anything else you want me to do before you merge it.

AgeManning (approved):

PR #97, opened 2022-01-13
Public kbuckets access and bucket index

ogenev:
In [Trin](https://github.com/ethereum/trin), we are manually managing the overlay instance of the routing table (`KBucketsTable<TNodeId, TVal: Eq>`) and the buckets (`Vec<KBucket<TNodeId, TVal>>`). For example, we are using a custom bucket refresh lookup algorithm which is first picking buckets with index 239-255, then we select a target bucket and generate a random NodeID falling in the target bucket range. (For more details please check this PR: https://github.com/ethereum/trin/pull/213)

To make this possible, we need public access to the `buckets`  field in `KBucketsTable` and a way to check if a random NodeIi falls in a target bucket index (mainly for testing purposes).

Please advise what would be the most convenient way to accomplish this, happy to tweak this PR to meet maintainers' preferences.

### DIFF
--- a/src/discv5.rs
+++ b/src/discv5.rs
@@ -283,6 +283,11 @@ impl Discv5 {
         self.local_enr.read().clone()
     }
 
+    /// Returns the routing table of the discv5 service
+    pub fn kbuckets(&self) -> KBucketsTable<NodeId, Enr> {
+        self.kbuckets.read().clone()
+    }
+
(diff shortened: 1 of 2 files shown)


AgeManning:
I dont see any problem with doing what you have here.

AgeManning (approved):

Issue #42, opened 2020-10-12
Combine active requests mapping into a struct


AgeManning:
There are now two active requests mapping and currently they are manually kept in sync with each other. 

Grouping these data structures will make upgrades more reliable and force the two mappings to always be in sync and prevent coding errors.

AgeManning:
To elaborate on this issue. 

For any given request (the actual object is not that important) we keep two mappings:
- https://github.com/sigp/discv5/blob/master/src/handler/mod.rs#L202
- https://github.com/sigp/discv5/blob/master/src/handler/mod.rs#L207

For some technical reasons, that I wont go into here, we sometimes need to search for requests via the `NodeAddress` or sometimes via the `MessageNonce`. We can use the latter mapping to get to the former as the latter maps `MessageNonce` to `NodeAddress`. 

The problem here is that these mappings must always exist and there should always be exactly one entry in each mapping per request. Because these are separate fields on a larger struct (the `Handler`), it is not enforced at any real level that these mappings always contain an entry per request. In fact there has been a serious issue in the past where these were not in sync.

To clarify, we can mutate one of the mappings (i.e remove an element) and not another in various parts of the code. It would be nice to ensure that this is not possible, so that we can be sure there is always an entry in both mappings for every request.

If we go through the code (handler/mod.rs), it should be possible to convince yourself, that every time we add or remove an element from each of these mappings their counter-part is added or removed. If this is not the case, there is a serious issue.

To make this code more robust and prevent the mappings from going out of sync in future code upgrades, we could introduce a new wrapper struct. 

This struct would hold both the mappings and introduce public functions that allow us to search via `NodeAddress` or `MessageNonce`. It would expose public functions that remove a request. When we remove a request, it removes it from both mappings internally. This way we can be sure that the internal mappings are always in sync. 

There are complications with this approach:
- One of the mappings is a `HashMapDelay` which implements `Stream`. I imagine we would have to implement `Stream` on the wrapper struct. This is because the requests can expire after a period of time. Polling the stream gives us the expired requests. 
- The current code inserts and removes elements from these mappings in different parts of the code, its not trivialy remove from both mappings at once. We'll have to look at how things are currently removed and inserted so that the main logic is generally followed if we now remove them simultaneously (i.e via a call to `remove()` on the wrapper struct, or simply by returning them in the Stream implementation). 


To be a little bit more concrete, we could have a wrapper struct something like:
```rust
struct ActiveRequests {
[...]

PR #101, opened 2022-03-25
Active requests mapping invariant

emhane:
## Issue Addressed

https://github.com/sigp/discv5/issues/42

## Proposed Changes

Wrap active_requests and active_requests_nonce mapping into one struct to handle as invariant.

## Additional Info

Nothing to add


divagant-martian:
Good job! Implementation looks solid. I did some additional tests that you don't need to do and they pass just fine.
Just for a matter of form I'm going to ask you a couple of changes:
1. Could you create a `src/handler/active_requests.rs` and move there the `ActiveRequests` struct along with it's impl blocks? This way the handler logic and the active requests mappings logic are each self contained. Easier to read and maintain
2. Could you `git rm src/handler/hashmap_delay.rs`? This is a remnant of before age moved it to a separate crate and was only used for the active requests

divagant-martian (commented):
Good job! Implementation looks solid. I did some additional tests that you don't need to do and they pass just fine.
Just for a matter of form I'm going to ask you a couple of changes:
1. Could you create a `src/handler/active_requests.rs` and move there the `ActiveRequests` struct along with it's impl blocks? This way the handler logic and the active requests mappings logic are each self contained. Easier to read and maintain
2. Could you `git rm src/handler/hashmap_delay.rs`? This is a remnant of before age moved it to a separate crate and was only used for the active requests

divagant-martian (approved):
lgmt

src/handler/mod.rs
```diff
@@ -580,36 +571,18 @@ impl Handler {
...
-            // Check for an active request
-            let node_address = match self.active_requests_nonce_mapping.remove(&request_nonce) {
-                Some(addr) => addr,
-                None => {
-                    trace!("Received a WHOAREYOU packet that references an unknown or expired request. Source {}, message_nonce {}", src_address, hex::encode(request_nonce));
+        let mut request_call = match self.active_requests.remove_by_nonce(&request_nonce) {
+            Some((node_address, request_call)) => {
+                // Verify that the src_addresses match
+                if node_address.socket_addr != src_address {
+                    trace!("Received a WHOAREYOU packet for a message with a non-expected source. Source {}, expected_source: {} message_nonce {}", src_address, node_address.socket_addr, hex::encode(request_nonce));
```
AgeManning, line 578: I think some logic has changed here. 

In this case, we want to add the mapping back. This will prevent other random people from tampering with peoples requests if they know the nonce etc.

AgeManning (commented):
Nice. Exactly what we were chasing. Code has improved a lot!

src/handler/mod.rs
emhane: Yes, I see it. Fixing it now.

PR #102, opened 2022-03-28
Active requests fixup

emhane:
## Issue Addressed

https://github.com/sigp/discv5/issues/42

## Proposed Changes

Fix security vulnerabilities of merged PR https://github.com/sigp/discv5/pull/101 

## Additional Info

Nothing to add

### DIFF
--- a/src/handler/active_requests.rs
+++ b/src/handler/active_requests.rs
@@ -38,7 +38,15 @@ impl ActiveRequests {
         match self.active_requests_nonce_mapping.remove(nonce) {
             Some(node_address) => match self.active_requests_mapping.remove(&node_address) {
                 Some(request_call) => Some((node_address, request_call)),
-                None => unreachable!("A matching request call mapping should always exist"),
+                None => {
+                    #[cfg(debug_assertions)]
+                    panic!("Panic on debug, a matching request call doesn't exist");
+                    #[cfg(not(debug_assertions))]
+                    {
+                        error!("A matching request call doesn't exist");
+                        return None;
+                    }
+                }
             },
             None => None,
         }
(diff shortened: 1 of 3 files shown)


src/handler/active_requests.rs
```diff
@@ -38,7 +38,12 @@ impl ActiveRequests {
         match self.active_requests_nonce_mapping.remove(nonce) {
             Some(node_address) => match self.active_requests_mapping.remove(&node_address) {
                 Some(request_call) => Some((node_address, request_call)),
-                None => None,
+                // If the matching request_call mapping doesn't exist, rollback remove_by_nonce op
+                None => {
+                    self.active_requests_nonce_mapping
```
AgeManning, line 43: In this case, I think it means we have an entry in `active_requests_nonce_mapping` but not one in `active_requests_mapping`. 

This is the situation we are trying to avoid and probably shouldn't add anything back.  I think in some cases the original code would panic in this scenario. Perhaps here we only panic in debug. So maybe in this case we put
```rust
debug_unreachable!()
```

src/handler/active_requests.rs
```diff
@@ -53,7 +58,12 @@ impl ActiveRequests {
                     .remove(request_call.packet.message_nonce())
                 {
                     Some(_) => Some(request_call),
-                    None => None,
+                    None => {
+                        // If the matching nonce mapping doesn't exist, rollback remove op
```
AgeManning, line 62: The same here

src/handler/active_requests.rs
emhane: ```debug_unreachable!()``` results in an unsafe code block, replaced the code that should be dead code with ```unreachable``` for now

src/handler/active_requests.rs
```diff
@@ -38,7 +38,7 @@ impl ActiveRequests {
         match self.active_requests_nonce_mapping.remove(nonce) {
             Some(node_address) => match self.active_requests_mapping.remove(&node_address) {
                 Some(request_call) => Some((node_address, request_call)),
-                None => None,
```
divagant-martian, line 41: It's better to add a message here for two reasons
1. It serves as a comment explaining why this should not happen
2. In the eventuality that it happens, the panic is going to print that message

src/handler/active_requests.rs
```diff
@@ -53,7 +53,7 @@ impl ActiveRequests {
                     .remove(request_call.packet.message_nonce())
                 {
                     Some(_) => Some(request_call),
-                    None => None,
```
divagant-martian, line 56: same here

divagant-martian (changes requested):
Thanks Emilia, I'll leave this to age to merge and we can decide about the unreachable

divagant-martian (approved):
lgtm, thanks :)

PR #105, opened 2022-04-04
force node contact to have a contactable node address

divagant-martian:
to include ipv6 support we need to be able to identify which socket address was used to contact a node (the ipv4 or ipv6 one). Before this was trivial because we assumed "sanitized" inputs in the handler structures that should all have an ipv4 address. This is an intermediate step to enforce the handler has contactable node contacts with explicit socket addresses. I'm opting for now for the simpler solution, but in the future I think we can optimize this

### DIFF
--- a/src/discv5.rs
+++ b/src/discv5.rs
@@ -25,6 +25,7 @@ use crate::{
 use enr::{CombinedKey, EnrError, EnrKey, NodeId};
 use parking_lot::RwLock;
 use std::{
+    convert::TryInto,
     future::Future,
     net::SocketAddr,
     sync::Arc,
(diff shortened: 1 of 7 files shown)


divagant-martian:
Now onto why I think this is the simplest, unoptimized option: We have duplication of data in various parts. The `NodeContact` itself contains the `public_key`/ `node id` twice. The optional `Enr` holds the rest of the data if present twice too. A `RequestCall` is mapped by a `NodeAddress`, and inside it contains a `NodeContact`, which has a `NodeAddress` itself. `pending_requests` in the handler is a map from `NodeAddress` to `ActiveRequest` which in turn has a `NodeContact`. The most useful optimization would be to avoid requiring the `Enr` in the `NodeContact` but I've yet to find how

divagant-martian:
closing this, since I'll include it in another PR in a better way

PR #109, opened 2022-04-14
Accept sessions when ENR doesn't match observed IP

carver:
There are some situations when a peer displays a mismatched IP/port from its ENR, and currently, those session requests are immediately rejected by `discv5`. I understand that the spec allows a more relaxed approach, and at least some other libraries take it. Some examples when ENR doesn't match:
- peer is unaware of NAT and doesn't have STUN to identify public IP
- peer is listening on 0.0.0.0, but must pick an IP for the ENR. When multiple IPs are available on the same machine, and two peers on the same machine choose different IPs and try to talk to each other, then the OS will route the sending packet through the closest IP to the destination, which won't match the sender's chosen IP.

Peers that are unaware of their NAT situation are leeching, at least temporarily. But it can be helpful to get the leeches connected and aware of their externally-visible IP addresses.

With this patch:
Do not populate the routing tables or propagate the ENR, but do respond to inbound requests.

### DIFF
--- a/src/handler/mod.rs
+++ b/src/handler/mod.rs
@@ -766,15 +766,27 @@ impl Handler {
                 Ok((session, enr)) => {
                     // Receiving an AuthResponse must give us an up-to-date view of the node ENR.
                     // Verify the ENR is valid
-                    if self.verify_enr(&enr, &node_address) {
-                        // Session is valid
-                        // Notify the application
+                    if enr.node_id() == node_address.node_id {
+                        // Signature is valid, but socket may not be
                         // The session established here are from WHOAREYOU packets that we sent.
                         // This occurs when a node established a connection with us.
-                        let _ = self
-                            .service_send
-                            .send(HandlerOut::Established(enr, ConnectionDirection::Incoming))
-                            .await;
+
+                        if self.verify_enr(&enr, &node_address) {
+                            // Notify the application that the ENR is fully validated
+                            let _ = self
+                                .service_send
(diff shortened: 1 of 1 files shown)


src/handler/mod.rs
```diff
@@ -766,15 +766,27 @@ impl Handler {
                 Ok((session, enr)) => {
                     // Receiving an AuthResponse must give us an up-to-date view of the node ENR.
                     // Verify the ENR is valid
-                    if self.verify_enr(&enr, &node_address) {
-                        // Session is valid
-                        // Notify the application
+                    if enr.node_id() == node_address.node_id {
+                        // Signature is valid, but socket may not be
```
carver, line 770: I didn't actually confirm in the source that the signature is validated before the node_id is set on `node_address`, I'm just hoping that's true. I could change this to simply:
```suggestion
                    if enr.node_id() == node_address.node_id {
                        // Node ID matches, but socket may not
```
Though at that point maybe I might as well remove the comment altogether...

carver:
I'd love thoughts from @AgeManning @divagant-martian and/or whomever else (not sure who to ping)

AgeManning:
This seems to come up quite a bit, I'll add a section to this on the readme. 

Here is the the reasoning behind it. Let me know if this doesn't address your concerns:
https://github.com/sigp/discv5/pull/70#discussion_r651505609

divagant-martian:
This seems to be a pitfall from the examples, in particular the `find_nodes` one, since given an ip we set it in the Enr, even if it's the unspecified address. We should probably update the example to minimize confusion

PR #110, opened 2022-04-15
Fix self-find query

### LINKED ISSUE #108: Unexpected "Requested a node find itself" error
I got the error when a response of FINDNODE contains ENR of the target node (the node passed to `Discv5::find_node()`).

## Error message

> 2022-04-14T01:11:36.349333Z ERROR discv5::service: Send RPC failed: Requested a node find itself

## Details of the case which occurs the issue

20220414_requested_a_node_find_itself

- (*1) https://github.com/sigp/discv5/blob/fdbb80432419578eebadf27c7d6fb098975f5de6/src/query_pool/peers/closest.rs#L180-L181
- (*2) https://github.com/sigp/discv5/blob/d47f7421f7d8a984cf0bceb1ec517310bb8dcf99/src/service/query_info.rs#L35-L36

## How to fix

I think we should skip to insert the peer into `closest_peers` if the peer received via NODES response and the target are same.

https://github.com/sigp/discv5/blob/fdbb80432419578eebadf27c7d6fb098975f5de6/src/query_pool/peers/closest.rs#L180-L181

I will file a PR to fix this issue. 😃

ackintosh:
## Issue Addressed

Resolves #108

## Proposed Changes

Skip incorporate the discovered peer into `closest_peers` if the peer is the target in order to avoid self-find query.

### DIFF
--- a/src/query_pool/peers/closest.rs
+++ b/src/query_pool/peers/closest.rs
@@ -175,6 +175,10 @@ where
 
         // Incorporate the reported closer peers into the query.
         for peer in closer_peers {
+            // Skip if the peer is the target in order to avoid self-find query.
+            if &peer == self.target_key.preimage() {
+                continue;
+            }
             let key: Key<TNodeId> = peer.into();
             let distance = self.target_key.distance(&key);
             let peer = QueryPeer::new(key, QueryPeerState::NotContacted);


AgeManning:
Hey @ackintosh 

Thank you very much for this :pray: 

I had a look through. The solution you propose here, means that we exclude the actual target from our queries. i.e if we search for a specific target and it is known, the results will not include that we found the actual target. 

I think the behaviour that we want is that if a node claims to know the actual target, we should then attempt to contact that target. If it is reachable, then we add it to the results. 

#117 is an attempt at solving this, where we still contact the target, and if its contactable return it with the results of the query. 

Could you test this PR and see if it resolves your issues also?

If you agree with me, perhaps we close this PR in favour of #117. 

Thanks again for finding this bug.

ackintosh:
> I think the behaviour that we want is that if a node claims to know the actual target, we should then attempt to contact that target. If it is reachable, then we add it to the results.

I understand the correct behaviour, thank you @AgeManning !

I'm testing #117 on [this test plan](https://github.com/ackintosh/test-plan-discv5/pull/13), will let  you know once the test has been finished.

ackintosh:
Closing this PR in favour of #117.

ackintosh:
I've finished the test, and have left [the comment](https://github.com/sigp/discv5/pull/117#issuecomment-1120700161).

PR #111, opened 2022-05-04
Enforce contactable nodes

divagant-martian:
change the node contact to keep track of the specific address used to contact a node


src/service.rs
```diff
@@ -531,20 +531,23 @@ impl Service {
...
                         }
                     }
                     // don't know of the ENR, request the update
                     _ => {}
                 }
                 if let Some(enr) = to_request_enr {
-                    self.request_enr(enr, None);
+                    // TODO: check this
+                    let contact =
+                        NodeContact::try_from_enr(enr).expect("stored ENRs are contactable.");
```
AgeManning, line 549: We probably should err on the side of caution and just log an error, rather than panicing here

src/service.rs
```diff
@@ -594,21 +597,24 @@ impl Service {
...
-            if let Ok(request_node_address) = active_request.contact.node_address() {
-                if request_node_address != node_address {
-                    warn!("Received a response from an unexpected address. Expected {}, received {}, request_id {}", request_node_address, node_address, id);
-                    return;
-                }
+
+            let expected_node_address = active_request.contact.node_address();
+            if expected_node_address != node_address {
+                warn!("Received a response from an unexpected address. Expected {}, received {}, request_id {}", expected_node_address, node_address, id);
+                // TODO: here we removed the request. Check it
```
AgeManning, line 604: Yes good point. The active_requests mapping heavily relies on the correctness of the handler. The handler also does this check, but in addition handles the timeouts for requests. 
I think in principle, if the handler works correctly, this is unreachable. But if we do reach this situation, we probably should fail loudly and we want to remove the mapping to avoid the mapping potentially growing indefinately

src/service.rs
```diff
@@ -831,16 +825,18 @@ impl Service {
...
+                enr_seq: self.local_enr.read().seq(),
+            };
+            let active_request = ActiveRequest {
+                contact,
+                request_body,
+                query_id: None,
+                callback: None,
+            };
+            self.send_rpc_request(active_request);
+        }
```
AgeManning, line 839: Maybe we log a warn here, so we can see if there are non-contactable nodes in our routing table. It shouldn't be the case, but being noisy will help identify them

src/service.rs
```diff
@@ -1026,8 +1022,14 @@ impl Service {
     ) {
         // find the ENR associated with the query
         if let Some(enr) = self.find_enr(&return_peer) {
+            let contact = match NodeContact::try_from_enr(enr) {
+                Ok(contact) => contact,
+                Err(NonContactable) => {
+                    return error!("Query {} has a non contactable enr", *query_id)
```
AgeManning, line 1028: I think we have to be careful here. 

A query is a larger group of RPC requests. I think in the previous version, we would send the request to the handler, the handler would find the ENR to be non-contactable, return a RequestError then we would mark the query request as failed and the query would be able to progress (its a big state machine)

If we are circumventing the handler here, I think we need to tell the query that this RPC request failed, so it can progress. 

This logic is complex and convoluted. Perhaps we should design it in a way that we can identify these kinds of things at compile time

AgeManning (commented):
This is a great improvement. 

The old design wasted resources sending contacts to the handler to determine if they were contactable. It's nicer to decide this on creation. 

I'll add to this PR and let you review my changes.

AgeManning:
:thinking: - For some reason I can't push to your branch. So I've made this: https://github.com/sigp/discv5/tree/contactable-node

AgeManning:
urgh.... you had a branch called `enforce-contactable-node` which is what autocomplete defaults to :rofl: :sob:

divagant-martian:
lol yeah sorry

src/service.rs
divagant-martian: done

src/service.rs
divagant-martian: added a loud debug fail and upgraded the log to err, not sure if you want to make it unreachable completely

src/service.rs
divagant-martian: I'm leaving this one to you since I haven't studied this part of the code yet and I don't find it feasible to study it soon / fast enough

PR #116, opened 2022-05-05
Ipv6 support

divagant-martian:
Continuation of #107 (see https://github.com/sigp/discv5/pull/107#issuecomment-1118820216 )


src/ipmode.rs
```diff
@@ -0,0 +1,240 @@
+use crate::Enr;
+use std::net::SocketAddr;
+///! A set of configuration parameters to tune the discovery protocol.
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum IpMode {
+    Ip4,
+    Ip6 { enable_mapped_addresses: bool },
+}
```
AgeManning, line 9: I think we should discuss a third variant - dual stack. 
As I understand, these are mutually exclusive and we might partition the network. A dual stack node might be able to bridge the two, but it will be a bit more work to add this in.

src/service.rs
```diff
@@ -780,23 +778,58 @@ impl Service {
...
+
                         if let Some(ref mut ip_votes) = self.ip_votes {
                             ip_votes.insert(node_id, socket);
-                            if let Some(majority_socket) = ip_votes.majority() {
-                                if Some(majority_socket) != local_socket {
-                                    info!("Local UDP socket updated to: {}", majority_socket);
-                                    self.send_event(Discv5Event::SocketUpdated(majority_socket));
-                                    // Update the UDP socket
-                                    if self
+                            let (maybe_ip4_majority, maybe_ip6_majority) = ip_votes.majority();
```
AgeManning, line 790: There is some interesting logic here. We are not checking our `IpMode`. So if we are in Ipv4 mode, we could update our ENR to include ipv6 if enough nodes tell us this is what they are contacting us on. 

Initially, I thought we wouldn't want to do this, but perhaps we do. We could be behind a NAT that is contacting via ipv6 and sending us internally ipv4 packets. This kind of setup, is going to be tricky as to what we allow in our routing table.

Probably should have a chat about the kind of logic we're going to support here (and make a doc). If we are in ipv4 mode, should we advertise ipv6 and vice-versa?
If we allow ipv4-only and ipv6-only to advertise their counter parts, then what should each allow in our local routing table. If we are dual stack, should we preference ipv6? If so, there could be an attack where people force an ipv4-only node to advertise an ipv6 address (erroneously), which will make them instantly non-contactable by nodes that prefer ipv6 when both are advertised.

Maybe the simplest approach is to not support networks where nodes are ipv4 local but NAT'd behind ipv6. Curious about your thoughts here.

src/socket/mod.rs
```diff
@@ -43,18 +50,41 @@ impl Socket {
...
+                        "Cannot create an ipv6 socket from an ipv4 address",
+                    )),
+                    SocketAddr::V6(ip6) => Ok((*ip6).into()),
+                }?;
+                let socket = Socket2::new(Domain::IPV6, Type::DGRAM, Some(Protocol::UDP))?;
+                let only_v6 = !enable_mapped_addresses;
+                socket.set_only_v6(only_v6)?;
+                socket.set_nonblocking(true)?;
+                socket.bind(&addr)?;
+                tokio::net::UdpSocket::from_std(socket.into())
```
AgeManning, line 79: I had issues doing this a while ago. 
Using a std socket caused some really funky issues and the solution was to not map from a std_socket. I've not looked into how to do this otherwise, or if its still an issue. 
See this: 
https://github.com/tokio-rs/tokio/issues/2969
and this:
https://github.com/tokio-rs/tokio/issues/3001

AgeManning (commented):
Nice. Making my way through. Some logic things that are prob worth discussing at some point. :)

src/ipmode.rs
divagant-martian: Age, dual stack is ipv6 with mapped addresses. There are 3 variants

src/socket/mod.rs
divagant-martian: The trick is the set non blocking. I haven't had any issues with that setup

src/service.rs
divagant-martian: So basically the thing is that the type of socket we use, and therefor the type of nodes we are able to send packets to, it's in principle not related to the address in which we are contactable. So for example, with this, now an ipv4 only node would be able to advertise itself to ipv6 (including dual stack) via tunneling

src/service.rs
```diff
@@ -1302,7 +1335,7 @@ impl Service {
     /// session key-pair has been negotiated.
     fn inject_session_established(&mut self, enr: Enr, direction: ConnectionDirection) {
         // Ignore sessions with non-contactable ENRs
-        if enr.udp4_socket().is_none() {
+        if self.config.ip_mode.get_contactable_addr(&enr).is_none() {
```
AgeManning, line 1338: this

PR #117, opened 2022-05-06
Gracefully handle queries that contain the target

AgeManning:
Queries that contain a target attempt to contact the target to search for itself.

Previously this resulted in an RPC error as highlighted in #108. 

This PR permits these RPC requests and adds the target in the found closest peers, if that peer is contactable.

### DIFF
--- a/src/discv5/test.rs
+++ b/src/discv5/test.rs
@@ -243,7 +243,7 @@ fn find_seed_linear_topology() {
     }
 
     for (node, previous_node) in main_result.iter().skip(1).zip(main_result.clone()) {
-        let key: kbucket::Key<NodeId> = node.clone().into();
+        let key: kbucket::Key<NodeId> = (*node).into();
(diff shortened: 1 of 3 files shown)


ackintosh:
I've tested this PR with [this test plan](https://github.com/ackintosh/test-plan-discv5/pull/13), and confirmed that the issue I've highlighted in #108 has been resolved!

--- 

However, while running [the test plan](https://github.com/ackintosh/test-plan-discv5/pull/13), I'm facing the other issue which is detailed on [this comment](https://github.com/ackintosh/test-plan-discv5/pull/13#issuecomment-1120430861).

> https://github.com/ackintosh/test-plan-discv5/pull/13#issuecomment-1120430861
> Race condition issue

I'm thinking of how to solve that, but no idea yet. 🤔 

Should I post a new issue for that? I would appreciate it if you let me know if you have any concerns about that. 🙏

PR #121, opened 2022-05-11
Handle overflow in `session.counter` explicitly

ackintosh:
https://github.com/sigp/discv5/blob/63f00ae56a912e4fc3aef8f0d910c9fee9a65f12/src/handler/session.rs#L60

This doesn't panic in release mode, but it would be better to use explicit `wrapping_*` operations.

### DIFF
--- a/src/handler/session.rs
+++ b/src/handler/session.rs
@@ -57,7 +57,7 @@ impl Session {
         src_id: NodeId,
         message: &[u8],
     ) -> Result<Packet, Discv5Error> {
-        self.counter += 1;
+        self.counter = self.counter.wrapping_add(1);
 
         // If the message nonce length is ever set below 4 bytes this will explode. The packet
         // size constants shouldn't be modified.
--- a/src/handler/session.rs
+++ b/src/handler/session.rs
@@ -251,3 +251,44 @@ impl Session {
         Ok((packet, session))
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::handler::session::{Keys, Session};
+    use crate::handler::{crypto, NodeContact};
+    use crate::packet::ChallengeData;
+    use enr::{CombinedKey, EnrBuilder, NodeId};
+    use std::convert::TryFrom;
+
+    #[test]
+    fn counter_overflow() {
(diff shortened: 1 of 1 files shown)


AgeManning:
Is this something you have observed to occur?

There is a lot of messages that need to happen before we overflow this thing. 

If it concerns you, happy to add this wrapping add. In theory, we shouldn't re-use counters (degrades security) but I think its unrealistic that in one session we would send this many messages, so happy to explicitly wrap. 

Do you think its fine to add as-is @divagant-martian - or would you prefer it explicitly for debug builds?

divagant-martian:
Also agree on not reusing counters, which we indeed do in theory in release builds if this ever happens, but at least failing in debug builds (current code) gives us a chance to identify a potential issue. If the idea is to improve code quality I would argue checked ops would be better

AgeManning:
Yeah agreed. 

Maybe in release we should even emit regular warnings indicating that we are re-using counters. Although I'm not sure we really need to cover the case. It will only happen when ~4.3 billion messages are sent to a single node on a single session. 
Sessions expire when not used also and a new session starts the counter again from scratch. 

So maybe we use checked ops and emit an error when we wrap. We can't increase the size of the counter as we're bound by the specs. 

What do you think @ackintosh?

ackintosh:
I've not observed any errors about this counter, the intent of this PR is to improve code quality.

I was thinking:

- 1 . Using explicit operations is recommended in the RFC
  - https://github.com/rust-lang/rfcs/blob/master/text/0560-integer-overflow.md#lint
    - > It would be better to use explicit checked or wrapped operations as appropriate.
  - The above is about unsafe code though.
- 2 . Basically I think it is not good that software behaves differently depending on what the mode of build, for troubleshooting
  - In release mode, overflow is specified to wrap. So I added `wrapping_add` in this PR.

--- 

Anyway, I agree that using checked ops. That is an ideal handling for the counter I think. 💡

ackintosh:
I changed my mind. 🙇 

As you said it is unrealistic, and even if counter reuse occurs,  it does not necessarily lead to vulnerability as the nonce is constructed with counter + random bytes.

Thank you @divagant-martian @AgeManning 🙏

PR #122, opened 2022-05-12
Correct Handshake race condition

AgeManning:
Attempt to handle issues raised by @ackintosh here: https://github.com/ackintosh/test-plan-discv5/pull/13#issuecomment-1120430861

From what I can tell it seems we attempt new handshakes even when one is underway. 

To address this, I think we can prevent sending messages if we are awaiting a challenge response. We wait for either the response, or a timeout before sending the message and optionally establishing a new session. 

This way, it should work the first go, and we don't have to increment the number of retries.

@ackintosh - Could you check to see if this PR resolves your issue. (Sorry for the delay, have a few things going on).

### DIFF
--- a/src/handler/mod.rs
+++ b/src/handler/mod.rs
@@ -24,7 +24,7 @@
 //!
 //! Requests from the application layer can be made via the receive channel using a [`HandlerIn`].
 //! Responses from the application layer can be made via the receive channel using a [`HandlerIn`].
-//! Messages from the a node on the network come by [`Socket`] and get the form of a [`HandlerOut`]
+//! Messages from a node on the network come by [`Socket`] and get the form of a [`HandlerOut`]
 //! and can be forwarded to the application layer via the send channel.
 use crate::{
     config::Discv5Config,
@@ -540,14 +540,11 @@ impl Handler {
             return;
         }
 
-        // Ignore this request if the session is already established
-        if self.sessions.get(&node_address).is_some() {
-            trace!(
-                "Session already established. WHOAREYOU not sent to {}",
-                node_address
-            );
-            return;
-        }
+        // NOTE: We do not check if we have an active session here. This was checked before
+        // requesting the ENR from the service. It could be the case we have established a session
+        // in the meantime, we allow this challenge to establish a second session in the event this
+        // race occurs. The nodes will decide amongst themselves which session keys to use (the
+        // most recent).
 
         // It could be the case we have sent an ENR with an active request, however we consider
         // these independent as this is in response to an unknown packet. If the ENR it not in our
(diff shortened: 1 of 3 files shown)


ackintosh:
I've tested with this PR, have observed the issue was happened in the sequence below. 

test-plan_pr13_request-timed-out_with_retry_2

Memo for myself: Testground logs
https://gist.github.com/ackintosh/9d63470429a5e35e23840e24758b5f10

AgeManning:
@ackintosh - Thanks for this. 

Thought a bit more about it and I think we still want to respond to all messages and there's always going to be some race condition if we start setting up a handshake at the same time. 

My current solution is to hold at most two pairs of session keys per session. I've managed to do this without affecting the memory footprint because we were already allocating this size. 

I'm hoping we don't oscillate between keys, I think this solution should be stable that once a pair of nodes fight over initial set of keys they will eventually agree on a set after a few rounds of communication. 

When you get a chance, could you check that this PR now addresses the issue?

ackintosh:
Tested with this PR, the frequency of  the issue has been reduced considerably. Occasionally the following issues occurred.

test-plan_pr13_request-timed-out_with_retry_3

Testground logs: https://gist.github.com/ackintosh/ed33aaeb63ecad7f11f18d454c1c28f8

AgeManning:
Ok cool. Sounds like we're getting closer.

Originally I hadn't planned on allowing two session keys co-exist, so I was preventing these late WHOAREYOUs. In the new context I think it makes sense to send them and allow the second handshake to complete.

I've made that modification. :crossed_fingers: its the last one and all your tests work. ;)

src/handler/mod.rs
```diff
@@ -552,6 +557,7 @@ impl Handler {
         let challenge_data = ChallengeData::try_from(packet.authenticated_data().as_slice())
             .expect("Must be the correct challenge size");
         debug!("Sending WHOAREYOU to {}", node_address);
+        self.add_expected_response(node_address.socket_addr);
```
divagant-martian, line 560: I can't right now tell if this is something that we were missing for an already existing `remove_expected_response` call, or if the call is missing after this change

divagant-martian, line 560: I can't right now tell if this is something that we were missing for an already existing `remove_expected_response` call, or if the call is missing after this change

AgeManning: I think it was missing before hand and I added it in as I found it.

There was also another bug where we could send a pending request but if it failed we were not reporting back to the service. 

Both of those fixes are in this PR I think.

AgeManning:
Thanks for the review @diva. 

I think we've resolved the issues we set out to. Let me know if you think its all good to merge

divagant-martian (approved):

PR #127, opened 2022-06-21
ENRs for Discv5

AgeManning:
This adds a brief outline for our design decision to drop requests from implementations that advertise non-contactable addresses in their ENR.

This issue continually comes up, so I think its necessary we add it to our README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,3 +73,54 @@ A simple example of creating this service is as follows:
       println!("Found nodes: {:?}", found_nodes);
    });
 ```
+
+# Addresses in ENRs 
+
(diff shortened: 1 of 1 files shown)


divagant-martian:
We have the private addresses filter in lighthouse but not here, so `127.0.0.1` is technically still contactable (we even use it in the examples) so we probably should discuss this only in the case in which the peer advertises a private address but it's actually sending via a public address. At the end, local discv5 deployments work anyway

AgeManning:
Agreed, I should be more careful in my wording. Will update

PR #134, opened 2022-07-21
WIP: NAT traversal via Rendezvous protocol

emhane:


AgeManning (commented):
Looking good so far, nice work

src/service.rs
```diff
@@ -366,8 +474,46 @@ impl Service {
...
+                        HandlerOut::RequestTimedOut(request_id, node_id) => {
+                            if CHECK_VERSION(&self.local_enr.read(), vec![NAT]) {
+                                // Drop the request and attempt establishing the connection to the peer via the
+                                // NAT traversal protocol for sending future requests to the peer, if this peer
+                                // was forwarded to us in a NODES response and we hence have a relay for it.
+                                if self.active_requests.get(&request_id).is_some() {
+                                    let local_enr = self.local_enr.read().clone();
+                                    if let Some(relays) = self.relays.get(&node_id) {
+                                        if let Some(contact_relay) = relays.iter().next() {
+                                            trace!("Requests to peer, that we have learnt of in some NODES response, keep timing out. Peer may be behind an asymmetric NAT. Trying to connect to peer via the NAT traversal protocol. Peer: {}, Relay: {}", node_id, contact_relay);
```
divagant-martian, line 486: While logging this is a bit too long. You could leave out the first part as a comment and as log just the end
```suggestion
                                            trace!("Trying to connect to peer via relay. Peer: {node_id}, Relay: {contact_relay}");
```

src/service.rs
```diff
@@ -366,8 +474,46 @@ impl Service {
                                 }
                             }
                         }
+                        HandlerOut::RequestTimedOut(request_id, node_id) => {
```
divagant-martian, line 477: the logic of this match branch is long enough to be better moved to its own function

src/service.rs
```diff
@@ -366,8 +474,46 @@ impl Service {
...
+                                // was forwarded to us in a NODES response and we hence have a relay for it.
+                                if self.active_requests.get(&request_id).is_some() {
+                                    let local_enr = self.local_enr.read().clone();
+                                    if let Some(relays) = self.relays.get(&node_id) {
+                                        if let Some(contact_relay) = relays.iter().next() {
+                                            trace!("Requests to peer, that we have learnt of in some NODES response, keep timing out. Peer may be behind an asymmetric NAT. Trying to connect to peer via the NAT traversal protocol. Peer: {}, Relay: {}", node_id, contact_relay);
+                                            self.send_relay_request(contact_relay.clone(), local_enr, node_id);
+                                        }
+                                    } else {
+                                        warn!("Request {} to peer {} timed out the max retry times, but we have no relays for the peer to attempt contacting it via the NAT traversal protocol.", request_id, node_id);
```
divagant-martian, line 490: this one is also a bit too long for logs

src/service.rs
```diff
@@ -366,8 +474,46 @@ impl Service {
...
+                                        relays.retain(|relay| relay.socket_addr() != rendezvous);
+                                        if let Some(contact_relay) = relays.iter().next() {
+                                            return Some(contact_relay.clone());
+                                        } else {
+                                            warn!("No further relays stored for peer (receiver) {}", receiver);
+                                        }
+                                    }
+                                    None
+                                };
+                                if let Some(relay) = next_relay(self) {
```
divagant-martian, line 511: You could use this directly since it's not needed anywhere else
```suggestion
                                let maybe_next_relay = self.relays.get_mut(&receiver).map(|relays| {
                                        // Only give each rendezvous one chance per receiver to relay.
                                        relays.retain(|relay| relay.socket_addr() != rendezvous);
                                        relays.iter().next().cloned()
                                    });

                                if let Some(relay) = maybe_next_relay(self).flatten() {
```

divagant-martian (commented):
partial review here, will continue tomorrow

src/discv5.rs
```diff
@@ -63,8 +91,19 @@ pub enum Discv5Event {
...
+    /// this connection has been assigned given remote port.
+    SessionEstablishedNatSymmetric(Enr, IpAddr, u16),
     /// Our local ENR IP address has been updated.
     SocketUpdated(SocketAddr),
+    /// Our local ENR NAT address and udp port has been updated indicating
+    /// this node is behind a NAT and is externally reachable on this address.
+    NATUpdated(SocketAddr),
+    /// Our local ENR NAT address has been updated, no single port could be found by ip voting
+    /// indicating this node is behind a symmetirc NAT and is externally reachable on this address.
+    NATSymmetricUpdated(IpAddr),
```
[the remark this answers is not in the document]
emhane: Of course, this will break backwards compatibility for clients deciding not to support the NAT version, I see it now. Thanks!

src/discv5.rs
```diff
@@ -63,8 +91,19 @@ pub enum Discv5Event {
...
+    /// A new NAT session has been established with a node behind a NAT.
+    SessionEstablishedNat(Enr, SocketAddr),
+    /// A new NAT session has been established with a node behind a symmetric NAT,
+    /// this connection has been assigned given remote port.
+    SessionEstablishedNatSymmetric(Enr, IpAddr, u16),
     /// Our local ENR IP address has been updated.
     SocketUpdated(SocketAddr),
+    /// Our local ENR NAT address and udp port has been updated indicating
+    /// this node is behind a NAT and is externally reachable on this address.
+    NATUpdated(SocketAddr),
```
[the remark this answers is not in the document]
emhane: I will leave this info to logging and leave the Discv5 Events unchanged for backwards compatibility

examples/find_nodes.rs
```diff
@@ -196,6 +196,10 @@ async fn main() {
                     Discv5Event::SessionEstablished(enr, _) => info!("Session established {}", enr),
                     Discv5Event::SocketUpdated(addr) => info!("Socket updated {}", addr),
                     Discv5Event::TalkRequest(_) => info!("Talk request received"),
+                    Discv5Event::NATUpdated(addr) => info!("Node is behind NAT, externally reachable address updated {}", addr),
+                    Discv5Event::NATSymmetricUpdated(ip) => info!("Node is behind a symmetric NAT, externally reachable ip updated {}", ip),
```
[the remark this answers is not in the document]
emhane: Yes that's right. But it is useful for us to keep the peer in our kbuckets, so that we can send requests to it. Then we can keep the rule "only ENRs with a contactable address in kbuckets" and that way we can also limit the number of symmetrically nat:ed peers based on the ENR.

src/handler/mod.rs
```diff
@@ -503,6 +559,20 @@ impl Handler {
         self.add_expected_response(node_address.socket_addr);
         self.send(node_address.clone(), packet).await;
 
+        if let RequestBody::RelayRequest {
+            ref from_node_enr, ..
+        } = call.request.body
+        {
+            // If this node is the initiator of the RELAYREQUEST, wait double the request time.
```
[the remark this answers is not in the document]
emhane: Awesome, thanks, this really helps me to improve my way of thinking around code structure.

src/service.rs
```diff
@@ -192,6 +201,85 @@ pub struct Service {
...
+    attempts: u8,
+}
+
+#[derive(Default)]
+struct AwaitingContactableEnr {
+    peers: HashMap<NodeId, NonContactableEnr>,
+}
+
+impl AwaitingContactableEnr {
+    fn request_enr(&mut self, node_id: &NodeId) -> Result<Option<Enr>, String> {
```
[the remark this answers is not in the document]
emhane: Why?

src/service.rs
```diff
@@ -192,6 +201,85 @@ pub struct Service {
...
+        }
+        Ok(None)
+    }
+
+    fn awaiting(&self, node_id: &NodeId) -> bool {
+        if let Some(peer) = self.peers.get(node_id) {
+            if peer.attempts >= MAX_REQEUST_ENR_ATTEMPTS {
+                return false;
+            }
+            return true;
```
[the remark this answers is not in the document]
emhane: What do you mean that the `unwrap_or_default` could be called on?

src/service.rs
emhane: Good point, that's the right way to think about closures, will I conditionally call this code in more than one place.

src/service.rs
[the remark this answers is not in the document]
emhane: Oh I see now, this is meant for the review under. Gotcha.

src/service.rs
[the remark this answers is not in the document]
divagant-martian: because Strings allocate in the heap, its a small efficiency improvement

src/discv5.rs
```diff
@@ -143,6 +197,62 @@ impl Discv5 {
...
+            // This NAT traversal protocol will only work for routers that set up the same port
+            // externally (WAN facing) as this machine is configured to use for discv5 on LAN.
+            // The protocol does not cover routers that choose a random external port to map to
+            // this machine's listen port.
+            match listen_socket {
+                SocketAddr::V4(socket4) => {
+                    if let Err(e) = self
+                        .local_enr
+                        .write()
+                        .set_udp4(socket4.port(), &self.enr_key.read())
```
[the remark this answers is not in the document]
emhane: Do you recall what this would have broke? The builder pattern? Or you mean the program would have crashed? I have drive-by removed the code.

emhane:
All that I still want to do with my NAT PR is:

- guard entrypoints to use NAT traversal protocol with supports_feature method, like "when a request times out...do we try to contact it via relay"
- clean up state bloat, some structures need a MAX number, like max relays to store per node, and a deadline when they expire.

And of course address any reviews you leave.

AgeManning:
I know you are AFK at the moment. As I go through, I'm making some changes. I'll try to make these into small self-contained commits and comment here as I make changes. This will give you the ability to revert and see what I've modified. 

I've changed the configuration parameters. I wanted to avoid multiple states that are invalid, i.e setting a symmetric_nat_limit but have no symmetric_nat peers allowed. I combined these parameters into a single config parameter: `symmetric_nat_limit` which takes a `Option<usize>`. This way, if users want to remove symmetric nat peers, they just use `Some(0)` (the current default). If they want no bounds, they just set to `None`. This way there is no conflicting options users can set making it harder to get into confusing operating states. 

There was an `asymmetric_nat_limit` option in the builder which didn't do anything, so I removed it. Perhaps it will be added in the future. 
We should add these new fields into the debug impl also, which I forgot to do. 

I also added a `CombinedFilter` struct, so we can combine filters together without having to keep creating new concrete types. There's probably some more ergonomic ways of doing it, but this works at least for now. 

Looking good so far :)

AgeManning:
There was a bug in the symmetric nat filter logic. 

If we had reached the nat symmetric filter limit, no new ENRs would be able to be added to the routing table. The filter was not checking if the current ENR was behind a NAT or not. So ENRs that were not NAT'd would be treated as if they were and be filtered.

AgeManning:
I moved the EnrExtension trait to the root directory. I think it makes it easier to find and easier to access globally. I think other crates/libraries might want to use this extension (we might need it in lighthouse also). 
I also renamed it to `EnrNat` to make it more specific, as we have other Enr Extension traits elsewhere. For example `EnrEth2`.

AgeManning:
I wasn't sure about the ergonomics of how we were going to add features in the future. 
You had a `Feature` enum which represented a `u8`. You were doing binary operations to keep track of the bitfield, which is nice :). 
To support three features or something, I think we would then have to create variants like:
```rust
Feature {
 Nat: 1,
 NatAndA: 2,
 NatAndAAndB: 7,
 NatAndBbutNotA: 5
 // etc
 }
 ```
 Which doesn't seem very ergonomic to handle all the permutations. So I modified this enum to a struct called `FeatureBitfield`. This allows you to set and read elements on a bitfield. I think it makes it easier to use, add features and cleans up a bit of the code. Let me know what you think.

**Comment by AgeManning** [comment]
I unified the two struct we used for voting into a single one, hopefully to simplify some code. 

I added logic to keep track of recent connections that are incoming, because I thought there could be a case that we have a routing table full of external peers but we still are receiving incoming connections, so that would indicate we are not behind a NAT. 

I removed some logic about banning a node if they send us a relay and we decide we are not behind a NAT. I haven't got to the sending of the relay request yet, but assumed it could happen if we dont respond in time, or network outages people could think we are behind a NAT and I thought maybe we dont want to jump to conclusions in banning them straight up. 

Maybe we add it back in once I get up to the logic about the relay requests.

**Comment by AgeManning** [comment]
Ok this last update I'm not sure about. I may have missed some understanding or problem, so please correct me if I have. 

I simplified a bunch of code by removing the logic associated with awaiting for NAT'd peer to update their ENRs. Let me explain my understanding of why it is here and some reasoning for why potentially it could be removed. Please let me know if I'm off or have not thought it through enough. 

Reason its here:
At the moment we only decide if a peer should go into our routing table on it's initial connection. Any subsequent update to its ENR does not give it a second chance to get in there. In addition to this, we instantly reject any peer that is "non-contactable" from being added. The concern then is for honest peers that are behind a NAT when they first start up, will appear non-contactable but can potentially later become contactable after updating their ENR. We want to give these peers a second chance of getting into the routing table. 

Why we might be able to remove it:
There is a bit of additional logic to handle this edge case. The edge case specifically refers to peers that initially advertise non-contactable ENRs but later decide to change their IP/NAT fields. For honest nodes, this usually covers two main groups:
1. This is the default state for all nodes starting up. They have `None` in their IP fields, until they find they are contactable and update it. 
2. The new NAT'd peers may find out they are behind a NAT and update their ENR accordingly. 

Before this PR, the first group existed and there is quite a few of them. Depending on how fast they update their ENR (they only really need 10 PONG's), they will have a smallish number of sessions to peers, lets say around 100-200. As these sessions are established, these nodes will not be entered into the routing table of their peers, because they are non-contactable. Later, they will update their ENR, and all their new connections give them the ability to be added into the routing table. Typically nodes maintain up to 5000 sessions at any given time. Once in someone's table, queries will spread them around everywhere, to many others tables.

In addition to this, most of the common buckets in our local routing table are full and often new connections don't make it into any bucket anyway. Also sessions expire eventually, and connections get re-established naturally giving second opportunities to get entered into the routing table.

For these reasons, we never added logic to handle the case that if a peer updates its ENR that we should give it a second chance immediately at entering the table. The reasons being specifically, 1) Because the bucket is likely full anyway, 2) Because they will get another attempt when their session expires naturally 3) Its only a small number of nodes (order 100) that it initially connected to before updating its ENR that have to worry about this and it will likely get in the table of some of the ~4000 other nodes it later connects to. 4) We have to handle a bunch of edge cases in the PING/PONG etc and this adds a bit to code complexity and possible DOS vectors etc getting us to download more ENR's than we might otherwise need, as well as memory in storing potential nodes ENRs that could be poorly configured (i.e not contactable).

I imagine the new group 2) behaves identically to the current group 1) so the reasons above apply. There may be some finer detail I'm missing with the NAT peers though. 

Anyway, curious about your thoughts on this.

File: src/service.rs [hunk 12]
```diff
@@ -610,6 +717,152 @@ impl Service {
...
+                                        "Updated local ENR's 'nat' and 'udp' field with socket {}",
+                                        socket
+                                    );
+                                    updated = true;
+                                }
+                                Err(e) => {
+                                    warn!("Failed to update local NAT socket. socket: {}, error: {:?}", socket, e);
+                                }
+                            }
+                        }
```
divagant-martian, line 768: Shouldn't we checking if we have nat support enabled here?

src/service.rs
```diff
@@ -197,8 +213,43 @@ pub struct Service {
...
 
+    /// A receiver/initiator peer that a hole-punch PING is sent to is stored to make sure only
+    /// one relay is used at a time when trying to hole-punch a NAT and to keep track of
+    /// connection direction in session establishment.
+    hole_punch_pings: HashSet<NodeId>,
+
     /// A channel that the service emits events on.
     event_stream: Option<mpsc::Sender<Discv5Event>>,
+
+    /// If this Discv5 instance is configured to allow peers behind symmetric NATs
```
divagant-martian, line 224: Why is this stored by distance? I couldn't find a strong use case for this

divagant-martian: if this is needed, it would be better to use the Log2Distance alias to make it clear

divagant-martian (commented):
I'm just halfway in the review but I have a couple of top level questions from the code:

AgeManning:
This last change is quite extensive. 

There were a few improvements I thought we could make. At a high level we were keeping track of symmetric connected ports in the handler and where indexing them by the bucket distance. It looked to me like this was being done in order to keep track of how many where in each bucket in order to not go over the limit. As Diva pointed out, we have a configuration and a filter to handle this, so it looked to me like that code could be removed entirely. 

In addition, we only need to keep track of the symmetric peer connected port for those peers that are in our routing table. I had considered moving it into the `Session` but we don't need it for every session as we wont form outbound connections with every session. The solution I ended up with, was creating a new variant to `NodeStatus` called `ConnectedSymmetricNat`. This means symmetric ports are stored in our routing table along with the ENR. If they leave the routing table, the ports are lost, so we don't need to worry about memory usage in keeping these mappings. Also its a single call to get the ENR and the status in one hit.

I also introduced a new type called `Connection`. This has allowed me to simplify a lot of code. It replaces the `RoutingType` in the Handler and allows me to unify all the `inject_session_x()` functions into a single one. 

I also had an attempt at grouping all the `try_from_enr_x()` and related functions, but proved more difficult than I anticipated.

In the end, I think I've managed to simplify a bunch of stuff, but maintain the logic.

src/handler/mod.rs
```diff
@@ -480,7 +467,26 @@ impl Handler {
...
+                        .send(HandlerOut::RequestFailed(
+                            request_id,
+                            RequestError::UnusedHolePunchPingToReceiver,
+                        ))
+                        .await
+                    {
+                        warn!("Failed to inform request failure {}", e)
+                    }
+                    return Ok(());
+                }
```
divagant-martian, line 489: I'm trying to figure out why is this done here. This function is `send_request`, which, when hole punching, completely ignores the request and instead informs an error?

divagant-martian: so if there is an ongoing challenge or request, this ping is queued as pending. It seems to be there is no need for that?

src/handler/mod.rs
```diff
@@ -853,7 +967,7 @@ impl Handler {
                 }
                 let id = request.id.clone();
                 trace!("Sending next awaiting message. Node: {}", contact);
-                if let Err(request_error) = self.send_request(contact, request).await {
+                if let Err(request_error) = self.send_request(contact, request, false).await {
```
divagant-martian, line 970: this request might have been queued with a hole punch = true. I think it either does not need to be queued or the values needs to be reused?

src/service.rs
```diff
@@ -1440,10 +1957,128 @@ impl Service {
...
+                | RequestError::TimedOutHolePunchPing => {
+                    // Don't update the connection if the request failed as expected according the
+                    // NAT traversal protocol. A RequestError::UnusedHolePunchPingToReceiver means
+                    // that a successful session has already been established. A
+                    // RequestError::TimedOutHolePunchPing usually happens if both nodes are
+                    // behind a NAT and this is the ping from the receiver to the initiator timing
+                    // out but setting the entry in its router's state table for the initiator's
+                    // hole punch ping coming next to enter.
+                    return;
+                }
```
divagant-martian, line 2010: `UnusedHolePunchPingToReceiver` looks unreachable to me.

There needs to be an active request to land here, but the `send_hole_punch_ping` fn does not add it if it's hole punch and this errors ir only returned in that case

src/service.rs
```diff
@@ -896,21 +1188,46 @@ impl Service {
...
+                self.send_hole_punch_ping(active_request);
+            } else {
+                _ = self.send_rpc_request(active_request);
             }
-            Err(NonContactable { enr }) => error!("Trying to ping a non-contactable peer {}", enr),
+        }
+    }
+
+    /// Sends a PING request to a node.
+    fn send_hole_punch_ping(&mut self, active_request: ActiveRequest) {
```
divagant-martian, line 1213: if this is just a ping, same as as with `send_ping` it should only receive the params needed for a ping. It's missleading being able to pass an `ActiveRequest` of any body type when this is not even going to be stored as an active request in `self.active_requests`

emhane:
> Ok this last update I'm not sure about. I may have missed some understanding or problem, so please correct me if I have.
> 
> I simplified a bunch of code by removing the logic associated with awaiting for NAT'd peer to update their ENRs. Let me explain my understanding of why it is here and some reasoning for why potentially it could be removed. Please let me know if I'm off or have not thought it through enough.
> 
> Reason its here: At the moment we only decide if a peer should go into our routing table on it's initial connection. Any subsequent update to its ENR does not give it a second chance to get in there. In addition to this, we instantly reject any peer that is "non-contactable" from being added. The concern then is for honest peers that are behind a NAT when they first start up, will appear non-contactable but can potentially later become contactable after updating their ENR. We want to give these peers a second chance of getting into the routing table.
> 
> Why we might be able to remove it: There is a bit of additional logic to handle this edge case. The edge case specifically refers to peers that initially advertise non-contactable ENRs but later decide to change their IP/NAT fields. For honest nodes, this usually covers two main groups:
> 
> 1. This is the default state for all nodes starting up. They have `None` in their IP fields, until they find they are contactable and update it.
> 2. The new NAT'd peers may find out they are behind a NAT and update their ENR accordingly.
> 
> Before this PR, the first group existed and there is quite a few of them. Depending on how fast they update their ENR (they only really need 10 PONG's), they will have a smallish number of sessions to peers, lets say around 100-200. As these sessions are established, these nodes will not be entered into the routing table of their peers, because they are non-contactable. Later, they will update their ENR, and all their new connections give them the ability to be added into the routing table. Typically nodes maintain up to 5000 sessions at any given time. Once in someone's table, queries will spread them around everywhere, to many others tables.
> 
> In addition to this, most of the common buckets in our local routing table are full and often new connections don't make it into any bucket anyway. Also sessions expire eventually, and connections get re-established naturally giving second opportunities to get entered into the routing table.
> 
> For these reasons, we never added logic to handle the case that if a peer updates its ENR that we should give it a second chance immediately at entering the table. The reasons being specifically, 1) Because the bucket is likely full anyway, 2) Because they will get another attempt when their session expires naturally 3) Its only a small number of nodes (order 100) that it initially connected to before updating its ENR that have to worry about this and it will likely get in the table of some of the ~4000 other nodes it later connects to. 4) We have to handle a bunch of edge cases in the PING/PONG etc and this adds a bit to code complexity and possible DOS vectors etc getting us to download more ENR's than we might otherwise need, as well as memory in storing potential nodes ENRs that could be poorly configured (i.e not contactable).
> 
> I imagine the new group 2) behaves identically to the current group 1) so the reasons above apply. There may be some finer detail I'm missing with the NAT peers though.
> 
> Anyway, curious about your thoughts on this.

I changed this because you told me too. Before I had code giving nodes a max number of tries to ping us with a contactable ENR.

emhane:
> I'm just halfway in the review but I have a couple of top level questions from the code:
> 
> **nat feature**: from the code it looks like a nat feature enabled/disabled node will turn on/off being a relay to other peers but will in any case update the nat fields using nat ip voting. I don't think there is a point in updating the nat ip vote results if the node does not support the nat (relay) feature, because upon receiving a relayrequest to itlsef it will reject it. With this, I think nat ip voting should be coupled with tha nat feature since it otherwise advertises a field that won't be of use to other peers
> 
> I thought about it like this: being a relay costs resources so if a node with limited resources can turn it off that's good, putting nodes in the KBuckets that are behind asymmetric NATs, as we do in v4, is a waste for the nodes that receive them in NODES responses from us, as those nodes won't be able to connect to them. If nodes that don't want to be relays still let other nodes know that they can only serve us as relaying our messages to the nodes they bootstrap with, that saves the network costs on a whole.
>
> **relay requets initiators** do we have a way of verifying a relayed relay request comes from the initiator node and that we are not being used to redirect traffic? This might be unnecessary since we are only sending pings and the resources used by the relayer and the receiver are esencially the same but I ask anyway
> 
> No. This would mean that we have to store all the NODES responses we send which is a huge state bloat. However I wrote code that bans nodes trying to get us to relay to a node that is not in our kbuckets, this would inhibit some repeated attempts to DoS certain nodes but not all. Age removed code that bans nodes for sending us faulty relay requests as mistakes happen, and a robust p2p network can survive without relying on bans, as I learnt. Only tested with max 20 nodes using my discv5-cli fork, would be cool to see in large scale.
>
> **number of relayed requests** Do we have a way to limit for how many peers are we a relay?
Nope but shouldn't be that hard to build. Possibly, a counter that increments when we receive a relay request and decrements on request timeout (or other failure) or when we relay the response.

src/handler/mod.rs
```diff
@@ -102,15 +109,90 @@ pub enum HandlerIn {
...
+    /// The [`ConnectionDirection`] is always [`ConnectionDirection::Incoming`] as peers behind
+    /// symmetric NATs only dial out. This peer will send PINGs to this node more frequently than
+    /// nodes that are not behind a NAT do, this way the hole punched in the peer's NAT for this
+    /// node will stay open and this peer can listen for requests from this node.
+    ///
+    /// Peers behind symmetric NATs cannot be hole-punched as they cannot not advertise a port for
+    /// new peers to hole-punch them on. The [`ConnectionDirection`] is always incoming as peers
+    /// behind symmetric NATs only dial out.
+    SymmetricNat(Enr, SocketAddr),
+}
```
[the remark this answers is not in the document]
emhane: No we don't because it has disappeared in some of @AgeManning commits!

The place where we really have to watch out is when doing the "am I behind a NAT?" voting and any variables that are set based on the peers correctly setting their ENR up till that point. Since if one incoming connection exists, and if this incoming connection is marked incoming based on the peers ENR, then we will believe we are behind a NAT.

src/service.rs
```diff
@@ -169,15 +184,16 @@ pub struct Service {
     /// All the iterative queries we are currently performing.
     queries: QueryPool<QueryInfo, NodeId, Enr>,
 
-    /// RPC requests that have been sent and are awaiting a response. Some requests are linked to a
+    /// RPC requests that have been sent and are awaiting a response. Some requests are linked to
```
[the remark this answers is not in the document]
emhane: Has already been fixed

src/service.rs
emhane: To check that we don't have more than `nat_symmetric_limit` peers behind a symmetric NAT per kbucket without intreating through all peers in a kbucket and checking their ENRs.

src/handler/mod.rs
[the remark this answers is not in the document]
emhane: Oh no I found it, it has been divided up into NatKind and ConnectionState in a new file called `src/connection.rs`

PR #141, opened 2022-10-20
Remove FINDNODE request bucket limit

jtraglia:
I don't see the purpose of this limit. If anything, I think it might hurt communications with other discovery implementations. It will ignore requests with more than 10 distances. The spec doesn't seem to have any limits on the number of distances in the request message, just the recommendation that the result is limited to 16 nodes, which is already done:

https://github.com/sigp/discv5/blob/878ef13378017bd8e5bf6426f23c3bf02f9e427d/src/service.rs#L996-L999

The `nodes_by_distance` function will stop adding to the results when the limit (`self.config.max_nodes_response`) is reached. By default, this configuration value is 16, [like the spec recommends](https://github.com/ethereum/devp2p/blob/master/discv5/discv5-wire.md#findnode-request-0x03). And the maximum packet size of 1280 should enforce some type of limit on the number of distances. Is there a reason for the limit of 10 distances?

### DIFF
--- a/src/kbucket.rs
+++ b/src/kbucket.rs
@@ -590,7 +590,7 @@ where
                 self.applied_pending.push_back(applied);
                 // Break if we've reached the maximum number of nodes we will provide in the
                 // response. There's no need to apply pending buckets past this point, the nodes
-                // won't be part of the response.
+                // in those buckets won't be part of the response.
                 node_count += bucket.num_entries();
                 if node_count >= max_nodes {
                     break
--- a/src/kbucket.rs
+++ b/src/kbucket.rs
@@ -593,7 +593,7 @@ where
                 // in those buckets won't be part of the response.
                 node_count += bucket.num_entries();
                 if node_count >= max_nodes {
-                    break
+                    break;
                 }
             }
         }
--- a/src/kbucket.rs
+++ b/src/kbucket.rs
@@ -581,12 +581,20 @@ where
             .collect::<Vec<_>>();
 
         // Apply pending nodes
+        let mut node_count = 0;
         for distance in &distances {
             // The log2 distance ranges from 1-256 and is always 1 more than the bucket index. For this
             // reason we subtract 1 from log2 distance to get the correct bucket index.
             let bucket = &mut self.buckets[(distance - 1) as usize];
             if let Some(applied) = bucket.apply_pending() {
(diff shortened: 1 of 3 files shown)


AgeManning:
I think potentially in an earlier version the `nodes_by_distances()` was a generic function and we would return a bunch of ENRs. I may have been worried about potential DOS vectors where malicious actors ask for all our buckets every request. 

I had another look over the code, at worst it makes us `apply_pending()` all our buckets on a request of the entire table but that is little over-head. 

So I don't see any issue with allow the entire table in requests. Only note would be to remove this outdated Note: https://github.com/sigp/discv5/blob/master/src/service.rs#L978

jtraglia:
> I think potentially in an earlier version the `nodes_by_distances()` was a generic function and we would return a bunch of ENRs. I may have been worried about potential DOS vectors where malicious actors ask for all our buckets every request.

Thanks, that makes sense. I'm guessing there wasn't a `max_nodes` response limit back then? 

> I had another look over the code, at worst it makes us `apply_pending()` all our buckets on a request of the entire table but that is little over-head.

How would you feel about moving that `apply_pending` call to this loop, at the top:

https://github.com/sigp/discv5/blob/cf35651a3a8340ded6d11cba058d909454f6e81a/src/kbucket.rs#L596-L617

That way, it would only commit buckets that will be returned. Keep in mind, distances are sorted and deduplicated.

AgeManning:
> How would you feel about moving that apply_pending call to this loop, at the top

If I recall correctly, I think a while ago I attempted to optimize this but resorted with two loops to correctly manage the mutable and immutable borrows. If they can be combined in a more efficient way, I'm down for it.

jtraglia:
Yeah, I soon realized why you did it that way. Instead of combining the loops like I originally had in mind, I've added a condition to the apply pending loop to break once it's reached the maximum number of nodes. How does this look?

AgeManning:
Yeah. The reason I remember this, is because everytime I look at this particular bit of code, I always forget and then try and combine the loops. I think I've tried like 5 times, haha. I think I've learnt now, to stop trying :). 

I think the `apply_pending()` isn't very expensive and we're splitting hairs (which is why I kinda left the two loops). But I like your approach. At least we wont be doing it to all our buckets.

AgeManning (approved):

PR #142, opened 2022-10-24
fix: start with bucket idx 0 if local key

mattsse:
If the `local_key` is provided for `ClosestBucketsIter` start it start in `Start(BucketIndex(0))` instead of `Done`. otherwise, the `closest_*` iterators will always be empty and make it impossible lookup the local node in the DHT.

this is also the state libp2p-kad starts with if the local key was provided.

https://github.com/libp2p/rust-libp2p/blob/e530118fb017122786180570b2b68f89aac53905/protocols/kad/src/kbucket.rs#L361-L365

See also: https://docs.rs/libp2p-kad/0.41.0/src/libp2p_kad/behaviour.rs.html#831-856

### DIFF
--- a/src/kbucket.rs
+++ b/src/kbucket.rs
@@ -767,7 +767,7 @@ impl ClosestBucketsIter {
     fn new(distance: Distance) -> Self {
         let state = match BucketIndex::new(&distance) {
             Some(i) => ClosestBucketsIterState::Start(i),
-            None => ClosestBucketsIterState::Done,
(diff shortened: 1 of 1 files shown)


AgeManning (approved):
Yeah nice, thanks. 

I've been meaning to add the standard kad bootstrap query as a general function in the `discv5` struct, just never got around to it, and hence why this bug has been left unnoticed.

Cheers :D

PR #143, opened 2022-10-27
Update max responses check for nodes response

jtraglia:
When reviewing the nodes response handler, I noticed that the maximum number of responses check is a bit weird. It will print a warning if there are going to be more than 15 packets (`DISTANCES_TO_REQUEST_PER_PEER * 5`) but it will stop processing after 6 (`self.config.max_nodes_response / 3 + 1`) packets. Am I understanding this correctly?

This PR will:

* Make a new constant: `MAX_RESPONSES`
* Use this constant for both checks.

### DIFF
--- a/src/service.rs
+++ b/src/service.rs
@@ -54,7 +54,8 @@ pub(crate) const DISTANCES_TO_REQUEST_PER_PEER: usize = 3;
 /// can be returned. Datagrams have a max size of 1280 and ENR's have a max size
 /// of 300 bytes. Bucket sizes should be 16. Therefore, to return all required peers
 /// there should be no more than `5 * DISTANCES_TO_REQUEST_PER_PEER` responses.
-pub(crate) const MAX_NODES_RESPONSES: usize = (MAX_NODES_PER_BUCKET / 4 + 1 ) * DISTANCES_TO_REQUEST_PER_PEER;
+pub(crate) const MAX_NODES_RESPONSES: usize =
+    (MAX_NODES_PER_BUCKET / 4 + 1) * DISTANCES_TO_REQUEST_PER_PEER;
 
 /// Request type for Protocols using `TalkReq` message.
 ///
@@ -733,7 +734,7 @@ impl Service {
                         // another response
                         // If we have already received all our required nodes, drop any extra
                         // rpc messages.
-                        if  current_response.received_nodes.len() < self.config.max_nodes_response
+                        if current_response.received_nodes.len() < self.config.max_nodes_response
                             && (current_response.count as u64) < total
                             && current_response.count < MAX_NODES_RESPONSES
                         {
--- a/src/service.rs
+++ b/src/service.rs
@@ -22,7 +22,7 @@ use crate::{
     handler::{Handler, HandlerIn, HandlerOut},
     kbucket::{
         self, ConnectionDirection, ConnectionState, FailureReason, InsertResult, KBucketsTable,
-        NodeStatus, UpdateResult,
+        NodeStatus, UpdateResult, MAX_NODES_PER_BUCKET,
     },
     node_info::{NodeAddress, NodeContact, NonContactable},
     packet::MAX_PACKET_SIZE,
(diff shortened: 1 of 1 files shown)


AgeManning:
Yeah. 
Originally the warning was to check for anything blatantly obviously wrong with other's implementations. It did a good job finding people sending a single ENR per request. 

The second bit of code was to actually be defensive. Although 15 would be the theoretical limit (assuming you can fit 4 ENRs into a single response), I think I wanted to be more stringent with actually getting spammed with responses. 

I don't quite remember the logic, but I was assuming you could at least fit 3 ENRs per response. So to get to our desired 16, at a minimum, you should be able to do it in 6. 

But you can ask for sparse buckets and probably this is too stringent. 

I agree with your changes, I added one other check to make it more directly dependent on how many nodes we have received. 

Let me know what you think

jtraglia:
> I don't quite remember the logic, but I was assuming you could at least fit 3 ENRs per response. So to get to our desired 16, at a minimum, you should be able to do it in 6.

Ah I understand now, that makes sense.

> I added one other check to make it more directly dependent on how many nodes we have received.

This is a good check check to have! Thanks for adding it.

Also, after thinking about it a little more, is 15 the actual upper limit? Shouldn't it be 12?

```rust
/// The number of distances (buckets) we simultaneously request from each peer.
/// NOTE: This must not be larger than 127.
pub(crate) const DISTANCES_TO_REQUEST_PER_PEER: usize = 3;

/// The maximum size of an ENR in bytes.
pub(crate) const MAX_ENR_SIZE: usize = 300;

/// A bucket can have `MAX_NODES_PER_BUCKET` (16) nodes and we request
/// `DISTANCES_TO_REQUEST_PER_PEER` (3) buckets, so there could up to be
/// `16 * 3 = 48` nodes to return. Datagrams have a max size of
/// `MAX_PACKET_SIZE` (1280) and ENRs have a max size of `MAX_ENR_SIZE` (300)
/// bytes, so they should be able to include at least `1280 / 300 = 4` nodes per
/// response. A peer can fit all of the nodes into `48 / 4 = 12` responses.
pub(crate) const MAX_NODES_RESPONSES: usize =
    (MAX_NODES_PER_BUCKET * DISTANCES_TO_REQUEST_PER_PEER) / (MAX_PACKET_SIZE / MAX_ENR_SIZE);
```

AgeManning:
Yeah so this is/was my logic:
- In a request we only really want to return 16 per node. We can ask for many distances, because sometimes the buckets can be sparse and we might only get 1 or 2 nodes in each bucket. From memory I think a query from a single node should get 16 at most. If we get more than 16 we just stop asking or caring about the responses.
- Although a packet has a max size of 1280, it doesn't guarantee we can fit 4 ENRs in that. I didn't actually do the math, but at the worst case, lets say we have 4 ENRs at their max-size being 1200 bytes. The packet also requires the RLP encoding, the RPC types and the two layers of encryption. If its an auth-header response, thats more bytes. I didn't look at the auth-header case, but I assume thats the worst case in which it may need more than 80 bytes. So in some absolute worst cases we may not be able to fit 4 ENRs, so I was thinking we should always at least fit 3.
- If the worst case is 3, then to get 16 nodes, its 16/3 = 5. I think I added one for rounding. 
- The original case was checking for blatantly wrong implementations. If we have honest node for a few buckets and those buckets are full. Then the worst case would be that the node sends all the peers in all the buckets we request. Assuming the logic above says 5 requests per bucket, then at the VERY worst, the node should only send 5 responses per distance (we can probably be stricter here). I was seeing some implementations sending more than this, and hence a warning was being emitted so I could identify which nodes were doing wrong behaviour.

In summary:
I think there are two limits we still want:
1) Is the node malicious/faulty : Some upper bound, like 5 responses per bucket. If they are over this, node is malicious/faulty
2) How many ENRs per request do we want before we stop reading packets (also kinda DOS prevention): I think this number is <  the number that determines if they are malicious/faulty). 

For 1) I had 5*Distances requested. And for 2) I added this extra check for number of nodes received and still use 1) as an upper bound. 

I dont think we can just do `MAX_PACKET_SIZE/MAX_ENR_SIZE` because there's extra stuff that needs to go into a packet than strait encoded ENRs. And the extra stuff is variable based on what kind of packet we are getting.

jtraglia:
Thank you for the detailed response. I understand and agree with all of that. Sorry if I am being nit-picky. I think the way you've designed things is smart. I'm happy with the current state of this PR if you want to keep it.

AgeManning (approved):

PR #148, opened 2022-11-18
Move RequestCall into its own file for easier review and self-contain…

divagant-martian:
just testing if ci runs for age


divagant-martian:
ci works

PR #149, opened 2022-11-20
Move RequestCall into individual file

AgeManning:
Moves the request call struct into its own file to reduce excess logic in the handler


divagant-martian (approved):
Lgtm

PR #150, opened 2022-11-21
cleanup socket creation wrt ip mode

divagant-martian:
I think this is equivalent and a bit cleaner


AgeManning (approved):
Looks good to me

PR #151, opened 2022-11-24
Reduce some clones in predicate impls

divagant-martian:
Reduces some unnecessary clones in around the predicate related types


AgeManning (approved):
Looks good to me. Nice

PR #152, opened 2022-11-26
Separate handler requests from service requests

divagant-martian:
right now the handler does requests on itself to get the ENR of a multiaddr session. This makes it so that if this request fails/timeouts we don't report those failures to the service.

Notes:
- This will be needed to handle nat hole punching handler requests correctly
- Right now, the FINDNODE request for a multiaddr session can both fail and timeout without affecting the session itself. So it seems to be that the only purpose of requesting and obtaining the enr is adding a peer to our kbuckets on a `HandlerOut::Established` event. Not sure if this is intended but seemed notable to me.

### DIFF
--- a/src/handler/active_requests.rs
+++ b/src/handler/active_requests.rs
@@ -2,7 +2,7 @@ use super::*;
 use delay_map::HashMapDelay;
 use more_asserts::debug_unreachable;
 
-pub(crate) struct ActiveRequests {
+pub(super) struct ActiveRequests {
     /// A list of raw messages we are awaiting a response from the remote.
     active_requests_mapping: HashMapDelay<NodeAddress, RequestCall>,
     // WHOAREYOU messages do not include the source node id. We therefore maintain another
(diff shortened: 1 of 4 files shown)


src/handler/active_requests.rs
```diff
@@ -2,7 +2,7 @@ use super::*;
 use delay_map::HashMapDelay;
 use more_asserts::debug_unreachable;
 
-pub(crate) struct ActiveRequests {
+pub(super) struct ActiveRequests {
```
divagant-martian, line 5: this was required to avoid leaking the  type that models a request private to the handler

src/handler/request_call.rs
```diff
@@ -1,17 +1,21 @@
...
-    rpc::{Request, RequestId},
+    rpc::{Request, RequestBody},
 };
 
+use super::HandlerReqId;
+
 /// A request to a node that we are waiting for a response.
 #[derive(Debug)]
-pub(crate) struct RequestCall {
+pub(super) struct RequestCall {
```
divagant-martian, line 11: Also required to prevent leaking types private to the handler

AgeManning (approved):
Yeah I see the issue you've mentioned now. 

This is a clean way to keep track of the types explicitly.

PR #155, opened 2022-12-19
feat: add value_mut function

mattsse:
Add `PresentEntry::value_mut` function so the value can be updated


divagant-martian (approved):
lgtm

AgeManning (approved):
LGTM

PR #157, opened 2022-12-20
Support configuring a different protocol id

### LINKED ISSUE #154: Support configuring a different protocol id
### Background

I am working on [rust-waku](https://github.com/LNSD/rust-waku) and, more precisely, on the [33/WAKU2-DISCV5 RFC](https://rfc.vac.dev/spec/33/) implementation.

The reference implementation, the Nim implementation ([nwaku](https://github.com/waku-org/nwaku)), of the Waku discv5 node discovery mechanism uses a different protocol ID for convenience. This is `d5waku`.

### Description

Right now, in **v0.1.0**, the `PROTOCOL_ID` is hardcoded:

https://github.com/sigp/discv5/blob/v0.1.0/src/packet/mod.rs#L32-L33

To make _rust-waku_ compatible with the Waku reference implementation, I would like to add support to configure the protocol ID (similar to what _rust-libp2p_ does with the Gossipsub protocol).

### Acceptance criteria

- [ ] Add the `protocol_id` field to `Discv5Config`. The default value must be `"discv5"`
- [ ] Add a matching `protocol_id` function to the `Discv5ConfigBuilder`.
- [ ] Add a `protocol_id` field to `packet::{Packet, PacketHeader}`.
- [ ] Encapsulate the `packet::Packet` codec logic to support checking the message validity against non-constant (non-static) values.

LNSD:
To make _rust-waku_ compatible with the Waku reference implementation, I added support to this crate to configure the protocol ID.

This PR resolves #154 

- [x] Add the `protocol` field to `Discv5Config`.
  + The protocol ID and the protocol version values default to the specification values, `discv5` and `0x0001`, respectively.
- [x] Add a matching `protocol_id` function to the `Discv5ConfigBuilder`. 
  + The configuration builder panics if the provided protocol ID has a length different than 6 bytes.
- [x] Add a `protocol_id` and `protocol_version` fields to `packet::{Packet, PacketHeader}`.
- [x] ~~Encapsulate the `packet::Packet` codec logic to support checking the message validity against non-constant (non-static) values.~~
- [x] Add support for checking the message validity against the configured values.

### DIFF
--- a/src/handler/session.rs
+++ b/src/handler/session.rs
@@ -17,6 +17,8 @@ pub(crate) struct Keys {
 /// A Session containing the encryption/decryption keys. These are kept individually for a given
 /// node.
 pub(crate) struct Session {
+    /// The Discv5 protocol id and version.
+    protocol: (&'static str, u16),
     /// The current keys used to encrypt/decrypt messages.
     keys: Keys,
     /// If a new handshake is being established, the older keys are maintained as race
@@ -37,8 +39,9 @@ pub(crate) struct Session {
 }
 
 impl Session {
-    pub fn new(keys: Keys) -> Self {
+    pub fn new(protocol_id: (&'static str, u16), keys: Keys) -> Self {
         Session {
+            protocol: protocol_id,
             keys,
             old_keys: None,
             awaiting_enr: None,
@@ -71,7 +74,10 @@ impl Session {
 
         // the authenticated data is the IV concatenated with the packet header
         let iv: u128 = rand::random();
+        let (protocol_id, protocol_version) = self.protocol;
         let header = PacketHeader {
+            protocol_id,
+            protocol_version,
             message_nonce,
             kind: PacketKind::Message { src_id },
         };
@@ -130,7 +136,9 @@ impl Session {
     /// Generates session keys from an authentication header. If the IP of the ENR does not match the
     /// source IP address, we consider this session untrusted. The output returns a boolean which
     /// specifies if the Session is trusted or not.
+    #[allow(clippy::too_many_arguments)]
     pub(crate) fn establish_from_challenge(
+        protocol: (&'static str, u16),
         local_key: Arc<RwLock<CombinedKey>>,
         local_id: &NodeId,
         remote_id: &NodeId,
@@ -207,11 +215,12 @@ impl Session {
             (None, None) => unreachable!("Checked in the first match above"),
         };
 
-        Ok((Session::new(keys), session_enr))
+        Ok((Session::new(protocol, keys), session_enr))
     }
 
     /// Encrypts a message and produces an AuthMessage.
     pub(crate) fn encrypt_with_header(
+        protocol: (&'static str, u16),
         remote_contact: &NodeContact,
         local_key: Arc<RwLock<CombinedKey>>,
         updated_enr: Option<Enr>,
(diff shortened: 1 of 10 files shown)


AgeManning:
Hey sorry, have been off. We will get to this, this week :)

divagant-martian (changes requested):
As far as I can tell there is no need or want to have a runtime protocol id, so I suggest you use an `&'static str` instead since cloning a String for every `Packet` is very expensive

divagant-martian:
@LNSD I've approved the runs for this PR, would you mind checking the failing checks?

LNSD:
> @LNSD I've approved the runs for this PR, would you mind checking the failing checks?

I would have bet that I ran `rustfmt` before pushing the commit. But seems that I didn't. Sorry, now it should be ok.

src/handler/session.rs
```diff
@@ -130,7 +136,9 @@ impl Session {
     /// Generates session keys from an authentication header. If the IP of the ENR does not match the
     /// source IP address, we consider this session untrusted. The output returns a boolean which
     /// specifies if the Session is trusted or not.
+    #[allow(clippy::too_many_arguments)]
```
LNSD, line 139: While fixing the _clippy_ warnings/errors, this function failed because it has "too many arguments". Truly solving this may require deeper refactoring, so I opted to ignore the warning for this specific method.

LNSD:
I hope this time the _cargo fmt_ and the _clippy_ issues are fixed 😅

AgeManning (commented):
Sorry for the delay. 

These changes look fine to me. 

it's kind of annoying that we have to thread the protocol-id everywhere. I can't think of a better solution except perhaps a lazy_static global variable which will avoid carrying the protocol-id around. 

The one thing that may be an issue, is that it looks like in every session we now store these extra protocol bytes. The handler keeps track of recent sessions, 1000 by default. So the extra bytes stored in every session isn't ideal.

If we dont want to go the lazy_static route, I wonder if its reasonable to store the protocol-id bytes in the handler and then just pass it to the session functions when creating packets, rather than storing it in the session itself?

If you dont see this as being an issue, happy to just merge this now and we can revisit it later if it has noticeable impact for us.

divagant-martian:
The `lazy_static` / `OnceCell` I don't find that hackish tbh. At the end it fits what we want: initialize only once. I would very much prefer that path to the current one. Open to hear different opinions

AgeManning:
@LNSD

There is now #158. This approach seems better to me. In Lighthouse we store up to 5000 sessions and where possible I'd prefer to not duplicate the protocol-id this many times by storing it in the session.  Also it seems #158 is smaller in code changes so I'd also opt for simplicity there. 

If #158 solves your issue and allows you to progress, I'd opt for #158. 

Going forward however, it might be best to build specific codecs which can account for the protocol-id outside of a global variable. The original design was based on a libp2p-like structure and was originally designed as a libp2p-protocol PR. As it is now a stand-alone crate, that design can be relaxed, but we still want to keep the encoding/decoding in their own tasks separate from the "encryption" (handler) as best we can. 

Anyway, let me know what you think and we can get some merging down and make a new release.

LNSD:
> This approach seems better to me. In Lighthouse we store up to 5000 sessions and where possible I'd prefer to not duplicate the protocol-id this many times by storing it in the session.

It completely makes sense to me. I am ok with the other PR approach. When I said "hackish" I was comparing it to the "codec" approach.

Let's close this PR in favour of #158. Thanks @divagant-martian and @AgeManning 😁

PR #158, opened 2023-02-08
Custom protocol

divagant-martian:


AgeManning (approved):
This looks like a simpler approach to me.

AgeManning:
Yep. This is has not been merged yet as I understand @divagant-martian might be improving on this slightly. 

Will wait for her response before merging.

divagant-martian:
Yeah, have not found time for this yet. Let's merge this if by monday I have still not found time for it

divagant-martian:
I explored the other option which involved threading down a trait with associated constants. This change however looks good enough for us and does not pollute the code with a trait that is only necessary at the `PacketHeader` level , so we are opting for using the current solution

LNSD:
It works for me too. Thanks 🙏

divagant-martian:
@LNSD this change created some issues, please have a look at https://github.com/sigp/discv5/pull/163 and let me know if this works for you. The main difference is that the protocol id and version will be moved to compile time. From [this old comment](https://github.com/sigp/discv5/pull/157#pullrequestreview-1242766829) I assume this will still be a solution for your needs

PR #159, opened 2023-02-16
Update `enr` crate version and add `serde` feature

ogenev:
I think it will be nice if `discv5` imports the `enr` crate with the `serde` feature, This allows Serialization/Deserialization of the `discv5::Enr` type out of the box.


divagant-martian:
Since `serde` is not explicitly brought by `discv5` it would be better to add it as feature that turns on the enr feature, and that is off by default

divagant-martian (approved):

PR #163, opened 2023-02-20
Custom protocol id

divagant-martian:
Allow configuring the protocol id and version at compile time. Reverts #158


LNSD:
From here: https://github.com/sigp/discv5/pull/158#issuecomment-1437291330

> From https://github.com/sigp/discv5/pull/157#pullrequestreview-1242766829 I assume this will still be a solution for your needs.

Yes, I don't need runtime configuration. A priori, this approach also looks good to me 👍🏼

Thanks, @divagant-martian

AgeManning (commented):
Yeah, I think this is a decent compromise. 

Its really annoying we have to thread the trait everywhere just to change the protocol-id. 

As we said on the previous PRs, we will look to adjust some of the design to make changes like this a bit easier in the future.

AgeManning (approved):

PR #167, opened 2023-03-14
Update delay map dependency

AgeManning:
This simply updates the delay_map dependency


divagant-martian:
#168 should fix ci

divagant-martian (approved):

PR #172, opened 2023-04-22
Expose ``send_ping`` from service.rs for discv5.rs

KolbyML:
I made a pr which adds ``send_ping`` to discv5.rs the reason for this is I am implementing the missing calls for Trin's discv5 rpc and in order to implement ``discv5_ping`` I need to expose it from this library first.

### DIFF
--- a/src/service.rs
+++ b/src/service.rs
@@ -38,7 +38,13 @@ use futures::prelude::*;
 use more_asserts::debug_unreachable;
 use parking_lot::RwLock;
 use rpc::*;
-use std::{collections::HashMap, net::SocketAddr, sync::Arc, task::Poll, time::Instant};
+use std::{
+    collections::HashMap,
+    net::{IpAddr, SocketAddr},
+    sync::Arc,
+    task::Poll,
+    time::Instant,
+};
 use tokio::sync::{mpsc, oneshot};
 use tracing::{debug, error, info, trace, warn};
 
@@ -146,6 +152,8 @@ pub enum ServiceRequest {
         Vec<u8>,
         oneshot::Sender<Result<Vec<u8>, RequestError>>,
     ),
+    /// The PING discv5 RPC function.
+    Ping(Enr, Option<oneshot::Sender<Result<Pong, RequestError>>>),
     /// Sets up an event stream where the discv5 server will return various events such as
     /// discovered nodes as it traverses the DHT.
     RequestEventStream(oneshot::Sender<mpsc::Receiver<Discv5Event>>),
@@ -213,12 +221,24 @@ struct ActiveRequest {
     pub callback: Option<CallbackResponse>,
 }
 
+#[derive(Debug)]
+pub struct Pong {
+    /// The current ENR sequence number of the responder.
+    pub enr_seq: u64,
+    /// Our external IP address as observed by the responder.
+    pub ip: IpAddr,
+    /// Our external UDP port as observed by the responder.
+    pub port: u16,
+}
+
 /// The kinds of responses we can send back to the discv5 layer.
 pub enum CallbackResponse {
     /// A response to a requested ENR.
     Enr(oneshot::Sender<Result<Enr, RequestError>>),
     /// A response from a TALK request
     Talk(oneshot::Sender<Result<Vec<u8>, RequestError>>),
+    /// A response from a Pong request
+    Pong(oneshot::Sender<Result<Pong, RequestError>>),
 }
 
 /// For multiple responses to a FindNodes request, this keeps track of the request count
@@ -335,6 +355,9 @@ impl Service {
                         ServiceRequest::Talk(node_contact, protocol, request, callback) => {
                             self.talk_request(node_contact, protocol, request, callback);
                         }
+                        ServiceRequest::Ping(enr, callback) => {
+                            self.send_ping(enr, callback);
+                        }
                         ServiceRequest::RequestEventStream(callback) => {
                             // the channel size needs to be large to handle many discovered peers
                             // if we are reporting them on the event stream.
@@ -428,7 +451,7 @@ impl Service {
                     };
 
                     if let Some(enr) = enr {
-                        self.send_ping(enr);
+                        self.send_ping(enr, None);
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


AgeManning:
This seems reasonable to me. Although we are starting to expose some of the lower level discv5 RPC methods which wasn't the original design. 

I'm not sure about the trin specs and if you need further access to lower level RPC things. If so, it could get complicated adding a lot more things into the discv5 struct. 

Currently the design goes like:

discv5: High-level struct holding an API for users to perform high-level discovery actions
service: A kademelia based DHT which uses discv5 underneath to perform the standard kademlia lookups etc
handler: Handles sending/receiving high-level discv5 packets and handles all the encryption/decryption logic. This directly exposes all the discv5 RPC methods. 
send/recv: Handles the encoding/decoding. 

The idea was that anyone that just wanted discv5 logic, could instantiate a handler and perform all the to-spec requests and receive responses for discv5 RPC methods, without having to worry about encryption/decryption etc. They just go, send Ping, and get a response etc. 

Service is the layer on top which technically isn't spec'd entirely as discv5. There's a lot of logic in there that isn't in the spec and people can change it at will to meet their needs. 

I'm not sure how complicated the trin spec is, but if you need access to a lot of discv5 RPC methods/complicated logic it might be worth considering forking the service/discv5 structs and importing the `Handler` struct. The `Handler` would maintain all the discv5 spec logic and then you could customize the `service` and `discv5` to suit trin however you see fit. I only mention this because it could get difficult in the future if Trin requires a lot more complicated logic than just the dht that is sitting on top at the moment. It might also work nicer for you if you customised the service to handle the ping response for trin, rather than passing it further up and out through `discv5`.

KolbyML:
> I'm not sure how complicated the trin spec is, but if you need access to a lot of discv5 RPC methods/complicated logic it might be worth considering forking the service/discv5 structs and importing the `Handler` struct. The `Handler` would maintain all the discv5 spec logic and then you could customize the `service` and `discv5` to suit trin however you see fit. I only mention this because it could get difficult in the future if Trin requires a lot more complicated logic than just the dht that is sitting on top at the moment. It might also work nicer for you if you customised the service to handle the ping response for trin, rather than passing it further up and out through `discv5`.

We only need ``ping`` and a ``none recursive ``find_node`` for trin. You guys already expose find_node recursively through ``find_node`` and support non-recursive findnode through ``request_enr()`` (the only issue is instead of using a multiple Addr we want to use an enr, distance. Adding a non-recursive find_node is added in my other PR)

So we don't want to add complicated logic we only want to expose ping (and a non-recursive findnode function with enr/distance) like you guys already did with non-recursive find_node (which was done with ``request_enr`` but with multiaddr and presets distance to 0) which is a "lower level discv5 RPC method" already exposed?

AgeManning:
Yep I understand. I think you guys added `talkreq`. 

Ok cool. So if its just these two things that you need, happy to expose them. If we start needing a bunch of extra things we may need to split up some of the higher level logic to make it easier to maintain. 

I'll go through and review this today :)

src/service.rs
```diff
@@ -213,12 +221,25 @@ struct ActiveRequest {
     pub callback: Option<CallbackResponse>,
 }
 
+#[derive(Debug)]
+#[allow(dead_code)]
```
AgeManning, line 225: Why do we have `allow(dead_code) ` here? Its used to construct the response right?

KolbyML:
Thank you very much :)

AgeManning (commented):
Looks good. Nice

src/service.rs
KolbyML:
```
error: fields `enr_seq`, `ip`, and `port` are never read
   --> src\service.rs:227:5
    |
225 | pub struct Pong {
    |            ---- fields in this struct
226 |     /// The current ENR sequence number of the responder.
227 |     enr_seq: u64,
    |     ^^^^^^^
228 |     /// Our external IP address as observed by the responder.
229 |     ip: IpAddr,
    |     ^^
230 |     /// Our external UDP port as observed by the responder.
231 |     port: u16,
    |     ^^^^
    |
    = note: `Pong` has a derived impl for the trait `Debug`, but this is intentionally ignored during dead code analysis
    = note: `-D dead-code` implied by `-D warnings`

error: could not compile `discv5` due to previous error 
```

```let response = Pong { enr_seq, ip, port };```

It is used to construct the response, but when I remove that clippy complains there is dead code for some reason?

KolbyML: Oh I just realized I should make them public, I wonder why I didn't think of that.

AgeManning: Yeah if they are not public, then the data would be unusable in the app?

I assume then this is not tested in the app side where you need it?

KolbyML: It isn't tested yet no. When I make a pr for trin adding support for this command I will have to write a test for it there. I was planning to do that after this was merged. I still need to ask the others if we only pull crates from releases now or also commits. Cause if we only do it from releases now then I might have to wait a while.

AgeManning: Sure. You could just do quick tests against the branch. That would have found things like not being able to access the data you want and us having to do another PR to correct it later. 

At the moment I think this is fine. I was thinking of getting your two PRs in and doing a release. I've also updated to the latest ENR.

KolbyML: I will test that tomorrow after my math final then and let you know :)

AgeManning:
I have been thinking, along with your other PR. It looks like you want access to be able to do misc discv5 RPC calls. 

We could make this generic. Allow the discv5 to send a generic `Request` which gets sent to the handler and the generic response matching the Id gets sent back (without touching the service). 

This, I imagine will allow users to call arbitrary discv5 RPC. It would encapsulate the PING, FINDNODEs (your two PRs) the TALKREQ and also any future requests we implement into discv5. 

It would be cleaner also as we'd just have a single function to handle all the possible requests and not individual functions for each thing we need as it comes up. 

What do you think about this approach?

KolbyML:
As long as it gets the job done I am all for it 👍. Plus if the code is simpler/more maintainable that is a win win

KolbyML:
Also I forgot to ask do you want me to shoot my shot on what I think you mean? Or are you planning to implement that since it was your conceptual concept? I am fine with both just let me know!

AgeManning:
Yeah, i went in to do it quickly last week, but its a bit more complicated than I expected. 

The issue is that some of the RPC requests can take many responses. I'm trying to think what the best API is. Do we internally try and manage the multiple responses and return final result in a future, or do we give the user a Stream and they can decide when to drop the stream. 

The TALKREQ also has `Drop` implemented which provides a return value to the peer. 

Trying to think of the most ergonomic way to give the user control of these messages without making it too difficult for them. I'll try knock out an implementation today.

AgeManning (approved):

AgeManning:
I built a version, but its not great from a user perspective with generic types. The responses then have to get matched specifically each time. Manually adding these type of things to the discv5 struct allows us to have specific return types making it much nicer. 

I also remember there's really not that many low-level rpc functions. This PR with your other ones, pretty much exposes all of the used ones. So lets just do that.

emhane:
Why did you need to send an extra ping explicitly? **Where is that ENR likely to be learnt about that is passed as a param to `send_ping`? By other means that through discv5 peer discovery I suppose?**

Otherwise also here I ask you, what is the problem with the existing ping interval that updates ENRs and simply call [kbuckets](https://github.com/sigp/discv5/blob/d86707d79c1183b14b8cf31ef62a8a74ef9cd3e4/src/discv5.rs#L296-L299) and then find the latest version of the enr at trin level? This way the flow of instructions between the top level tasks service and handler that implement the core discv5 protocol isn't distorted either by the way trin uses it as `Discv5`, the app facing level, holds a shared reference to the kbuckets. If the problem is that the interval to update ENRs is too slow why do you not just set the [ping_interval](https://github.com/sigp/discv5/blob/715fc822dfd7f9d2fb1eb43cf4b8e771768c283b/src/config.rs#L256-L260) ? @KolbyML

KolbyML:
> Why not just call [`add_enr`](https://github.com/sigp/discv5/blob/715fc822dfd7f9d2fb1eb43cf4b8e771768c283b/src/discv5.rs#L184-L225) for that use case? Or send it a talk request and wait for the talk response? Why the ping? i.e. why does trin need the pong
> 
> https://github.com/sigp/discv5/blob/715fc822dfd7f9d2fb1eb43cf4b8e771768c283b/src/rpc.rs#L95-L103
> 
> @KolbyML

image

I am implementing RPC for discv5_*, not history_* (I think you might be confused about that cause your response to my other PR you were referring to history) I am confused why we would send a talk_request when we aren't doing anything on a sub-protocol. There is already a discv5_addEnr RPC. The specs lay out the ping discv5_ping request responds with PONG.

PR #184, opened 2023-05-19
Respond to a PING even if the source doesn't match the ENR

ackintosh:
## Description

Our discv5 implementation drops a request if the source IP address doesn't match the ENR. For PING request, due to this behaviour, the source node can't notice that its external IP address has been changed.

Here is a simulation that reproduce that issue. 
https://github.com/ackintosh/discv5-testground/pull/98

This PR makes discv5 to respond PING even if the source IP address doesn't match.

### one-time sessions

I introduced **one-time session** at handler to encrypt PONG response without storing a (normal) session. One-time session here is a session for a specific request such that a PING request mentioned above. As the name implies, one-time session is removed once it's used.

### New config parameters

Added config parameters for one-time session introduced by this PR.

- one_time_session_timeout
- one_time_session_cache_capacity

These are similar to the parameters for normal sessions, but the default values are set to 30 seconds, and 100 respectively.

### Simulation test for this PR

Also I have created a Testground test plan that simulates IP changes. 
https://github.com/ackintosh/discv5-testground/pull/98

The result is fine. Here is a sequence diagram that illustrates how this change behaves on IP changes.

```mermaid
sequenceDiagram
    participant Node1
    participant Node2 ... Node N

    rect rgb(10, 10, 10)
    Note left of Node1: They communicate with each other<br> to establish a session.
    Note over Node1,Node2 ... Node N: Session established
    end

    rect rgb(10, 10, 10)
    Note left of Node1: Node1 changes its IP address<br> but doesn't update its ENR.
    Note over Node1: *** Change its IP address ***
    end

    Node1 ->> Node2 ... Node N: PING
    Note over Node2 ... Node N: No session with the new IP exists
    Node2 ... Node N ->> Node1: WHOAREYOU
    Note over Node1: Establish new session
    Node1 ->> Node2 ... Node N: Auth message<br> (The ENR record has the old IP)
    Note over Node2 ... Node N: `verify_enr()` results in false<br> as the ENR and src IP don't match.
    Note over Node2 ... Node N: Establish one-time session with Node1's new IP
    
    rect rgb(0, 255, 0)
    Note over Node2 ... Node N: Respond the PING using the one-time session
    end

    Node2 ... Node N ->> Node1: PONG

    rect rgb(0, 255, 0)
    Note over Node1: Update local ENR to the new IP via the IpVote process.
    end

    Node1 ->> Node2 ... Node N: PING
    Note over Node2 ... Node N: No (normal) session with the new IP exists
    Node2 ... Node N ->> Node1: WHOAREYOU
    Note over Node1: Establish new session
    Node1 ->> Node2 ... Node N: Auth message<br> (The ENR record has the new IP)

    Note over Node2 ... Node N: `verify_enr()` results in true here
    rect rgb(0, 255, 0)
    Note over Node2 ... Node N: Establish new (normal) session
    end

    Node2 ... Node N ->> Node1: PONG
```

## Change checklist

- [x] Self-review
- [x] Tests if relevant

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -35,6 +35,12 @@ pub struct Discv5Config {
     /// The maximum number of established sessions to maintain. Default: 1000.
     pub session_cache_capacity: usize,
 
+    /// The one-time session timeout. Default: 30 seconds.
+    pub one_time_session_timeout: Duration,
+
+    /// The maximum number of established one-time sessions to maintain. Default: 100.
+    pub one_time_session_cache_capacity: usize,
+
     /// Updates the local ENR IP and port based on PONG responses from peers. Default: true.
     pub enr_update: bool,
 
@@ -127,6 +133,8 @@ impl Discv5ConfigBuilder {
             request_retries: 1,
             session_timeout: Duration::from_secs(86400),
             session_cache_capacity: 1000,
+            one_time_session_timeout: Duration::from_secs(30),
+            one_time_session_cache_capacity: 100,
             enr_update: true,
             max_nodes_response: 16,
             enr_peer_update_min: 10,
@@ -196,6 +204,18 @@ impl Discv5ConfigBuilder {
     /// The maximum number of established sessions to maintain.
     pub fn session_cache_capacity(&mut self, capacity: usize) -> &mut Self {
         self.config.session_cache_capacity = capacity;
+        self
+    }
+
+    /// The one-time session timeout.
+    pub fn one_time_session_timeout(&mut self, timeout: Duration) -> &mut Self {
+        self.config.one_time_session_timeout = timeout;
+        self
+    }
+
+    /// The maximum number of established one-time sessions to maintain.
+    pub fn one_time_session_cache_capacity(&mut self, capacity: usize) -> &mut Self {
+        self.config.one_time_session_cache_capacity = capacity;
         self
     }
 
@@ -325,6 +345,11 @@ impl std::fmt::Debug for Discv5Config {
             .field("request_retries", &self.request_retries)
             .field("session_timeout", &self.session_timeout)
             .field("session_cache_capacity", &self.session_cache_capacity)
+            .field("one_time_session_timeout", &self.one_time_session_timeout)
+            .field(
+                "one_time_session_cache_capacity",
+                &self.one_time_session_cache_capacity,
+            )
             .field("enr_update", &self.enr_update)
             .field("query_parallelism", &self.query_parallelism)
             .field("report_discovered_peers", &self.report_discovered_peers)
--- a/src/config.rs
+++ b/src/config.rs
@@ -35,12 +35,6 @@ pub struct Discv5Config {
     /// The maximum number of established sessions to maintain. Default: 1000.
     pub session_cache_capacity: usize,
 
-    /// The one-time session timeout. Default: 30 seconds.
-    pub one_time_session_timeout: Duration,
-
-    /// The maximum number of established one-time sessions to maintain. Default: 100.
-    pub one_time_session_cache_capacity: usize,
-
     /// Updates the local ENR IP and port based on PONG responses from peers. Default: true.
     pub enr_update: bool,
 
@@ -133,8 +127,6 @@ impl Discv5ConfigBuilder {
             request_retries: 1,
             session_timeout: Duration::from_secs(86400),
             session_cache_capacity: 1000,
-            one_time_session_timeout: Duration::from_secs(30),
-            one_time_session_cache_capacity: 100,
             enr_update: true,
             max_nodes_response: 16,
             enr_peer_update_min: 10,
@@ -204,18 +196,6 @@ impl Discv5ConfigBuilder {
     /// The maximum number of established sessions to maintain.
     pub fn session_cache_capacity(&mut self, capacity: usize) -> &mut Self {
         self.config.session_cache_capacity = capacity;
-        self
-    }
-
-    /// The one-time session timeout.
-    pub fn one_time_session_timeout(&mut self, timeout: Duration) -> &mut Self {
-        self.config.one_time_session_timeout = timeout;
-        self
-    }
-
-    /// The maximum number of established one-time sessions to maintain.
-    pub fn one_time_session_cache_capacity(&mut self, capacity: usize) -> &mut Self {
-        self.config.one_time_session_cache_capacity = capacity;
         self
     }
 
@@ -345,11 +325,6 @@ impl std::fmt::Debug for Discv5Config {
             .field("request_retries", &self.request_retries)
             .field("session_timeout", &self.session_timeout)
             .field("session_cache_capacity", &self.session_cache_capacity)
-            .field("one_time_session_timeout", &self.one_time_session_timeout)
-            .field(
-                "one_time_session_cache_capacity",
-                &self.one_time_session_cache_capacity,
-            )
             .field("enr_update", &self.enr_update)
             .field("query_parallelism", &self.query_parallelism)
             .field("report_discovered_peers", &self.report_discovered_peers)
--- a/src/handler/mod.rs
+++ b/src/handler/mod.rs
@@ -191,6 +191,8 @@ pub struct Handler {
     active_challenges: HashMapDelay<NodeAddress, Challenge>,
     /// Established sessions with peers.
     sessions: LruTimeCache<NodeAddress, Session>,
+    /// Established sessions with peers for a specific request.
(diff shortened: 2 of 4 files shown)


ackintosh:
With the changes so far, a node get respond PING even if the source IP doesn't match ENR record. However still we can not update our IP in ENR.


```mermaid
sequenceDiagram
    participant Node1
    participant Node2

    rect rgb(10, 10, 10)
    Note left of Node1: <Node1's KBucket><br>[Node2] direction:Outgoing, state:Connected
    end

    rect rgb(10, 10, 10)
    Note left of Node1: Node1 changes its IP address<br> but doesn't update its ENR.
    end 
    Note over Node1: *** Change its IP address ***

    Node1 ->> Node2: PING
    Note over Node2: No session with the new IP exists
    Node2 ->> Node1: WHOAREYOU
    Note over Node1: Establish new session
    rect rgb(255, 255, 0)
    Note left of Node1: *** The direction is updated from Outgoing to Incoming<br> when establishing a new session here. ***
    end
    rect rgb(255, 255, 0)
    Note left of Node1: <Node1's KBucket><br>[Node2] direction:***Incoming***, state:Connected
    end
    Node1 ->> Node2: Auth message
    Note over Node2: Establish one-time session with Node1's new IP
    Note over Node2: Respond the PING using the one-time session
    Node2 ->> Node1: PONG
    rect rgb(255, 255, 0)
    Note over Node1: IpVote process ignores the PONG<br> since the connection direction with Node2 is Incoming.
    end
```

ackintosh:
To address the above [issue](https://github.com/sigp/discv5/pull/184#issuecomment-1562032339), I have created another PR. This PR modifies the handler to give ConnectionDirectionInstruction to the service instead of ConnectionDirection so that we can keep an existing connection direction.

https://github.com/ackintosh/discv5/pull/1

src/handler/mod.rs
```diff
@@ -813,6 +828,37 @@ impl Handler {
                         // established. If so process them.
                         self.send_next_request::<P>(node_address).await;
                     } else {
+                        // Respond to PING request even if the ENR or NodeAddress don't match
```
divagant-martian, line 831: Here if we store just one one-time-session per peer, we would want to check first for one such session being already established before doing the decode

divagant-martian (commented):
Couple suggestions about the logic and some about the code. Ty for the great work as always

src/handler/mod.rs
```diff
@@ -515,17 +519,28 @@ impl Handler {
...
-                    return;
-                }
-            };
-            self.send(node_address, packet).await;
+            send!(session);
+        } else if let Some(mut session) = self
+            .one_time_sessions
+            .remove(&(node_address.clone(), response.id.clone()))
+        {
+            send!(session);
```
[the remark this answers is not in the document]
ackintosh: Yeah I like it!  Readable and cleaner code is always appreciated. Thank you. 🙂

src/handler/mod.rs
```diff
@@ -191,6 +191,8 @@ pub struct Handler {
     active_challenges: HashMapDelay<NodeAddress, Challenge>,
     /// Established sessions with peers.
     sessions: LruTimeCache<NodeAddress, Session>,
+    /// Established sessions with peers for a specific request.
+    one_time_sessions: LruTimeCache<(NodeAddress, RequestId), Session>,
```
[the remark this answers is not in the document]
ackintosh: You're right. That was just my mistake. 😅  I was thinking of storing only one one-time session per node, as you said.

src/handler/mod.rs
ackintosh: > we would want to check first for one such session being already established

You mean we would want to move the second statement of below (`else if ... self.remove_one_time_session() ...`) to the top of the if statement?

https://github.com/sigp/discv5/blob/fce0fce685ad7aae4193124123996f85b7f30beb/src/handler/mod.rs#L522-L535

divagant-martian: yep, this is what I mean

ackintosh: My slight concern about doing that is, the second statement (`if ... self.remove_one_time_session() ...`) is false in most cases. Establishing a one-time session is under limited conditions, so the result of the second statement should be false in most send_response() calls.

Do you have a good reason for moving the second statement to the first?

divagant-martian: My reasoning is that is not impossible to have a one-time session created for a peer, then a normal session created, and then hit this line. A bit of a race condition. Moving the check up would prevent this. It's a matter of correctness but I understand your pov as a matter of performance. What are your thoughts on this?

ackintosh: I see your point. Your suggestion is right in terms of correctness. 

> a one-time session created for a peer, then a normal session created, and then hit this line

Yeah it looks a bit of a race condition but I think that if a normal session created, it's fine to use the normal session even if a one-time session (for the request) exists since the normal session is valid as well. The one-time session will be evicted over time without being used.

ackintosh:
@AgeManning @divagant-martian This PR is ready for review now.

src/config.rs
```diff
@@ -35,6 +35,12 @@ pub struct Discv5Config {
     /// The maximum number of established sessions to maintain. Default: 1000.
     pub session_cache_capacity: usize,
 
+    /// The one-time session timeout. Default: 30 seconds.
+    pub one_time_session_timeout: Duration,
+
+    /// The maximum number of established one-time sessions to maintain. Default: 100.
+    pub one_time_session_cache_capacity: usize,
+
```
divagant-martian, line 43: Since these config values are user-facing, we should probably explain here/somewhere what one time sessions are and what they do

src/handler/mod.rs
```diff
@@ -820,6 +826,37 @@ impl Handler {
...
+                            );
+                            self.one_time_sessions
+                                .insert(node_address.clone(), (request.id.clone(), session));
+                            if let Err(e) = self
+                                .service_send
+                                .send(HandlerOut::Request(node_address, Box::new(request)))
+                                .await
+                            {
+                                warn!("Failed to report request to application {}", e)
+                            }
```
divagant-martian, line 858: this will self clean, but if failed to report the request, there is no point in storing the one-time-session

divagant-martian (commented):
minor things. Otherwise letting age finish the review

src/handler/mod.rs
```diff
@@ -820,6 +826,37 @@ impl Handler {
...
+                                Ok(Message::Request(request)) if request.msg_type() == 1 => {
+                                    Some(request)
+                                }
+                                _ => None,
+                            },
+                            _ => None,
+                        };
+                        if let Some(request) = maybe_ping_request {
+                            debug!(
+                                "Responding PING request using one-time session. node_address: {}",
```
AgeManning, line 847:
```suggestion
                                "Responding to a PING request using a one-time session. node_address: {}",
```

AgeManning (approved):
This looks good to me. It should make us a little more friendly to people adjusting their IPs.

src/config.rs
divagant-martian: yep good idea

src/handler/mod.rs
divagant-martian: @AgeManning do you have any opinion here?

src/handler/mod.rs
ackintosh: Thanks for pointing it out!

ackintosh:
Thank you guys for the review! I have made changes based on the feedback.

@AgeManning Please let us know if you have any thoughts on https://github.com/sigp/discv5/pull/184#discussion_r1234093965.

src/handler/mod.rs
AgeManning: Not really. 
Its a single hash lookup that we are worried about. I think if we have a session, we shouldn't establish a one-time session. But there is a race condition where a session gets created after. So I think its a relatively rare case that we have both sessions. In this case, we probably want to prioritise the normal session over the one-time session. 

Seeing as we are probably running this a bit, I agree with @ackintosh in that it's nicer to avoid the hash lookup (i.e check and remove the one-time session) to handle this edge case. It's also no real loss because the one-time sessions only last 30 seconds (assuming they get pruned). 

But imo we are debating over a single hash lookup vs correctness, so i'm really on the fence. 

Maybe just leave it as @ackintosh has it, because we don't have to do another commit? 
I don't feel strongly at all, if either one of you do, happy to go that way.

divagant-martian: also of no strong opinion so merged :shrug: seems we are all happy so

PR #188, opened 2023-06-04
Fix unexpected ConnectionDirection update

ackintosh:
## Description

A connection direction can unexpectedly change from Outgoing to Incoming in the scenario shown in [this diagram](https://github.com/sigp/discv5/pull/184#issuecomment-1562032339). The diagram shows a specific case in PING/PONG. However, if we generalize this, when an `Outgoing` connection is established and the other node returns WHOAREYOU, the ConnectionDirection will change to `Incoming` at the time the new session is established. This could potentially cause some problems. e.g. IpVote doesn't work as IpVote ignores Incoming connections.

This PR makes the service to set only initial connection direction and not update it if a node already exists in the routing table.

## Notes & open questions

I have tested this PR with the PING/PONG simulation (https://github.com/sigp/discv5/pull/184), and confirmed that the Outgoing connection directions are kept.

```mermaid
sequenceDiagram
    participant Node1
    participant Node2

    rect rgb(10, 10, 10)
    Note left of Node1: <Node1's KBucket><br>[Node2] direction:Outgoing, state:Connected
    end

    rect rgb(10, 10, 10)
    Note left of Node1: Node1 changes its IP address<br> but doesn't update its ENR.
    end 
    Note over Node1: *** Change its IP address ***

    Node1 ->> Node2: PING
    Note over Node2: No session with the new IP exists
    Node2 ->> Node1: WHOAREYOU
    Note over Node1: Establish new session
    rect rgb(0, 255, 0)
    Note left of Node1: The Outgoing direction is kept<br> when establishing a new session here. 
    end
    rect rgb(0, 255, 0)
    Note left of Node1: <Node1's KBucket><br>[Node2] direction:***Outgoing***, state:Connected
    end
    Node1 ->> Node2: Auth message
    Note over Node2: Establish one-time session with Node1's new IP
    Note over Node2: Respond the PING using the one-time session
    Node2 ->> Node1: PONG
```

## Change checklist

- [x] Self-review
- [x] Tests if relevant

### DIFF
--- a/src/handler/mod.rs
+++ b/src/handler/mod.rs
@@ -113,7 +113,7 @@ pub enum HandlerOut {
     /// A session is only considered established once we have received a signed ENR from the
     /// node and either the observed `SocketAddr` matches the one declared in the ENR or the
     /// ENR declares no `SocketAddr`.
-    Established(Enr, SocketAddr, ConnectionDirection),
+    Established(Enr, SocketAddr, ConnectionDirectionInstruction),
 
     /// A Request has been received from a node on the network.
     Request(NodeAddress, Box<Request>),
@@ -138,6 +138,17 @@ pub enum ConnectionDirection {
     Incoming,
     /// We contacted the node.
     Outgoing,
+}
+
+/// An instruction from the handler to the service how to update the connection direction.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum ConnectionDirectionInstruction {
+    /// Incoming.
+    Incoming,
+    /// Outgoing.
+    Outgoing,
+    /// If there is an established connection keep it; if not Incoming.
+    IncomingIfNotExists,
 }
 
 /// A reference for the application layer to send back when the handler requests any known
@@ -682,11 +693,11 @@ impl Handler {
                 // outgoing session is that we originally sent a RANDOM packet (signifying we did
                 // not have a session for a request) and the packet is not a PING (we are not
                 // trying to update an old session that may have expired.
-                let connection_direction = {
+                let connection_direction_instruction = {
                     match (request_call.initiating_session(), &request_call.body()) {
-                        (true, RequestBody::Ping { .. }) => ConnectionDirection::Incoming,
-                        (true, _) => ConnectionDirection::Outgoing,
-                        (false, _) => ConnectionDirection::Incoming,
+                        (true, RequestBody::Ping { .. }) => ConnectionDirectionInstruction::Incoming,
+                        (true, _) => ConnectionDirectionInstruction::Outgoing,
+                        (false, _) => ConnectionDirectionInstruction::IncomingIfNotExists,
                     }
                 };
 
@@ -705,7 +716,7 @@ impl Handler {
                     .send(HandlerOut::Established(
                         enr,
                         node_address.socket_addr,
-                        connection_direction,
+                        connection_direction_instruction,
                     ))
                     .await
                     .unwrap_or_else(|e| warn!("Error with sending channel: {}", e));
@@ -795,7 +806,7 @@ impl Handler {
                             .send(HandlerOut::Established(
                                 enr,
                                 node_address.socket_addr,
-                                ConnectionDirection::Incoming,
+                                ConnectionDirectionInstruction::Incoming,
                             ))
                             .await
                         {
@@ -976,7 +987,7 @@ impl Handler {
                                                 .send(HandlerOut::Established(
                                                     enr,
                                                     node_address.socket_addr,
-                                                    ConnectionDirection::Outgoing,
+                                                    ConnectionDirectionInstruction::Outgoing,
                                                 ))
                                                 .await
                                             {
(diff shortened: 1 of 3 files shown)


ackintosh:
@AgeManning @divagant-martian Please have a look when you have time.

AgeManning (commented):
On a first review, it seems like we could leave the handler as it is, then in the service, only ever update the routing table if an entry doesn't exist?

This way the code changes would be minimal, and the initial connections are kept as final (until the node is exited from the routing table). 

What do you think?

src/handler/mod.rs
```diff
@@ -682,11 +693,11 @@ impl Handler {
...
                 // trying to update an old session that may have expired.
-                let connection_direction = {
+                let connection_update = {
                     match (request_call.initiating_session(), &request_call.body()) {
-                        (true, RequestBody::Ping { .. }) => ConnectionDirection::Incoming,
-                        (true, _) => ConnectionDirection::Outgoing,
-                        (false, _) => ConnectionDirection::Incoming,
+                        (true, RequestBody::Ping { .. }) => ConnectionUpdate::Incoming,
+                        (true, _) => ConnectionUpdate::Outgoing,
+                        (false, _) => ConnectionUpdate::IncomingIfNotExists,
```
divagant-martian, line 700: Checking this, I agree that the third match is a Incoming only if it doesn't exist, but it seems to be the first one is as well. I think the renaming is worth doing since it better explains what is going on `IncomingIfNotExists` is clearer than incoming in the sense that this is a conditional update.

This means then, removing the `Incoming` variant and adjusting accordingly

divagant-martian (commented):
I find this change in the right direction but I missed something on the first review that changes the point of the PR

src/handler/mod.rs
AgeManning: I have no real opinion on the name change. Happy to do it if we like. 

But I think we probably should also apply the same logic for outgoing connections. If there is already a node that is incoming, we probably don't want to promote it to outgoing. So I guess the options we should do would be:
1. Rename these varients, but probably also `OutgoingIfNotExists`
2. Leave it as it is, and just take the meaning to mean for new connections only. Change the service to only set the connection direction and not modify it (if an ENR already exists)

divagant-martian: talking with age, the suggestion would be to 
1. remove the conditional that checks the ping (first match statement), so that we depend only on checking if we are initiating the session or not
2. (this one is mine) keep the new names `ConnectionUpdate` definitely is an improvement, and if we want, keep the `*IfNotExists` (or perhaps `*IfNew`) variant names. I find they help with code clarity.
3. As consequence of 1) there would only be two connection updates, the "if new" logic would be entirely moved to the service (as you have already done) but applied to both connection directions
I find this a cleaner solution both in code clarity and cognitive load - goodbye ping special cases!

Can you guys give a thumbs up if we agree?

AgeManning: I agree. Just to be clear tho, in the changes we are never updating a connection. We only set the initial connection in the routing table and it now never gets updated. Perhaps the name: `ConnectionUpdate` could be a bit misleading. But it could just be me and I dont mind the name change at all. 
So this sounds good to me.

divagant-martian: I see your point. I think that us considering this as an update stems from this extra logic that's being moved to the service. In that case, agree on maintaining the original names

ackintosh: I agree with the suggestion. I'm also agree with maintaining the original names since in the suggestion we never update connection direction.

ackintosh: I have made changes to apply the suggestion and now am doing tests with the simulation locally, please give me a little time.

src/service.rs
```diff
@@ -1368,29 +1368,25 @@ impl Service {
...
-            ConnectionUpdate::Outgoing => ConnectionDirection::Outgoing,
-            ConnectionUpdate::IncomingIfNotExists => {
-                let key = kbucket::Key::from(node_id);
-                match self
-                    .kbuckets
-                    .read()
-                    .get_bucket(&key)
-                    .map(|bucket| bucket.get(&key))
-                {
-                    Some(Some(node)) if node.status.is_connected() => node.status.direction,
```
ackintosh: In this commit, I removed the `if node.status.is_connected()` condition because I think the condition gives attackers chances to promote their connection from Incoming to Outgoing.

ackintosh:
@AgeManning @divagant-martian I have tested again with the latest changes, and that was fine. 🙆‍♂️ Please have another look.

src/service.rs
```diff
@@ -1368,13 +1368,27 @@ impl Service {
...
+    fn inject_session_established(&mut self, enr: Enr, connection_direction: ConnectionDirection) {
         // Ignore sessions with non-contactable ENRs
         if self.ip_mode.get_contactable_addr(&enr).is_none() {
             return;
         }
 
         let node_id = enr.node_id();
+
+        // We never update connection direction if a node already exists in the routing table as we
+        // don't want to promote the direction like from incoming to outgoing.
```
divagant-martian, line 1380:
```suggestion
        // We never update connection direction if a node already exists in the routing table as we
        // don't want to promote the direction from incoming to outgoing.
```

divagant-martian (approved):
LGTM, missing a typo

AgeManning (approved):
Yeah nice, looks good. 

Much simpler :)

PR #189, opened 2023-06-05
Add some convenience functions for ListenConfig

AgeManning:
Adds some helping functions that make creating a ListenConfig a little bit nicer.


divagant-martian:
fmt is unhappy. Also, should we delete the "builder"-like functions? They seem way redundant with this (improved) additions

divagant-martian (approved):

PR #190, opened 2023-06-06
Improve logs for ipv6 support

AgeManning:
Downgraded an error log which would occur quite frequently for ipv6 only nodes in an ipv4-dominated network. 

I also increased a log which was hiding messages being dropped because of invalid source addresses.


divagant-martian (approved):
yep sounds good!

Issue #144, opened 2022-11-09
Speed up ENR update


AgeManning:
We do pings every 5 minutes after connections for security reasons. 

I think its probably safe for us to ping outgoing connections that end up in our routing table on initial connection.

divagant-martian:
can you be more specific when you say "on initial connection" ?

AgeManning:
Yep. We initially had that when we established a connection (more specifically, create a new session) we would send a PING. This has security implications in that malicious users could start connections with us and try to game the IP voting mechanism.

Now, we only ping connections every ping-interval, and only connections in the routing table. Although this is safer, the downside is that we don't start receiving PONG responses until after the ping-interval on start-up. This can be a decent amount of time.

Perhaps a middle ground is that new connections that are outgoing and make it into the routing table, we send PING's to when they first enter the table. This would allow us to start getting PONG responses early in order to update our ENR faster (when measuring time from start-up)

PR #196, opened 2023-06-18
Speed up ENR update

ackintosh:
## Description

closes https://github.com/sigp/discv5/issues/144.

By this PR, we send PING immediately when the session is established if the direction is `Outgoing`.

## Notes & open questions

I have tested this PR with the enr-update test plan. You can run the test plan locally.
https://github.com/ackintosh/discv5-testground/pull/109

Also I have tested with lighthouse (current unstable `Lighthouse/v4.2.0-cc780aa`). The ENR update had happened within 113sec avg from start-up.

```bash
Jun 25 05:33:06.794 INFO Lighthouse started, version: Lighthouse/v4.2.0-cc780aa, module: lighthouse:579
...
...
Jun 25 05:33:50.189 INFO Address updated, udp_port: 3365, ip: 133.32.132.191, service: libp2p, module: lighthouse_network::discovery:1012
```

## Change checklist

- [x] Self-review
- [x] Tests if relevant

### DIFF
--- a/src/service.rs
+++ b/src/service.rs
@@ -1270,17 +1270,18 @@ impl Service {
                     direction,
                 };
 
-                if direction == ConnectionDirection::Outgoing {
-                    // PING immediately if the direction is outgoing. This allows us to receive a PONG without
-                    // waiting for the ping_interval, making ENR updates faster.
-                    self.send_ping(enr.clone(), None);
-                }
-
                 match self.kbuckets.write().insert_or_update(&key, enr, status) {
                     InsertResult::Inserted => {
                         // We added this peer to the table
                         debug!("New connected node added to routing table: {}", node_id);
                         self.peers_to_ping.insert(node_id);
+
+                        // PING immediately if the direction is outgoing. This allows us to receive
+                        // a PONG without waiting for the ping_interval, making ENR updates faster.
+                        if direction == ConnectionDirection::Outgoing {
+                            self.send_ping(enr.clone(), None);
+                        }
+
                         let event = Discv5Event::NodeInserted {
                             node_id,
                             replaced: None,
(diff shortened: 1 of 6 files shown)


src/service.rs
```diff
@@ -1394,7 +1394,13 @@ impl Service {
...
             node_id, direction
         );
-        self.connection_updated(node_id, ConnectionStatus::Connected(enr, direction));
+        self.connection_updated(node_id, ConnectionStatus::Connected(enr.clone(), direction));
+
+        // PING immediately if the direction is outgoing. This allows us to receive a PONG without
+        // waiting for the ping_interval, making ENR updates faster.
+        if matches!(direction, ConnectionDirection::Outgoing) {
+            self.send_ping(enr, None);
+        }
```
divagant-martian, line 1403: looking to reduce the clone I suggest moving this to the `connection_updated` fn. This way the clone is only done when the dirction is outgoing instead of every time https://github.com/ackintosh/discv5/pull/2/files

ackintosh: I have merged your suggestion, thank you!

AgeManning:
Yeah nice. 

I think we should move this statement into the match just below it tho, so that we only ping peers that manage to make it into the routing table. 

The reason being, is I'm a little worried that if we do a FINDNODE to a malicious peer, that can direct to us to connect (in an outgoing direction) to a bunch of other malicious peers that want us to edit our IP address. 

If we limit this PING approach to only peers that make it into our routing table we get the added security measures that are applied there, like bucket and table limits which would help mitigate this scenario.

divagant-martian:
@ackintosh can you test this works as expected in testground? I see no reason to believe it won't but let's be sure. I'll merge after your confirmation :tada:

ackintosh:
Yeah, I was just testing this again. I have tested this change with Testground and Lighthouse, and confirmed that this works as expected. 🙆‍♂️

divagant-martian (approved):

Issue #198, opened 2023-06-30
Handling multiple in-flight requests


carver:
When sending multiple requests, the current design queues them, waiting for a response to the first request before sending the next one:

https://github.com/sigp/discv5/blob/a9ded0490229418cb61e747d55d3e0879b62e1d7/src/handler/mod.rs#L468-L481

If a client has a 100ms roundtrip time with a peer, that puts a hard upper limit of 10 messages per second with that peer. That limits the "bandwidth" to less than 12.8kB/s (102 kb/s).

There's nothing apparent in the protocol spec that requires 1 message at a time. Including the request ID even suggests that it is intended to support multiple in-flight requests.

What are maintainers' thoughts on reviewing a PR that would support multiple in-flight requests?

fjl:
It is certainly possible to run concurrent requests, but please note that it will make session handling much more complicated. Remember that in discv5, it is assumed that the session may be dropped at any moment. If the node has multiple requests in flight with a peer, and the peer answers with a challenge (WHOAREYOU), all pending requests would need to be replayed. It is much easier to implement the handshake flow when there is the guarantee that only a single request is pending for each peer.

AgeManning:
Yep. I think in its original design it allowed for multiple con-current in-flight requests. But handling all the edge cases got pretty unwieldy and there were significant code simplifications we can make by having a single one per-peer. 

The other issue you may run into, is that we also need a way to determine whether a peer is still alive/online. There is a configuration parameter that allows you to retry a request if it fails. If we have many outgoing requests at a time and some fail, should we consider the connection failed?

At the moment I think the default is that if we get a request time out, we just consider the peer offline and move on. This is helpful for discovery queries where we are asking many different peers lots of things at once, rather than a single peer a lot of things.

njgheorghita:
Took a stab at implementing this in #200 - would love to hear some feedback. Did I take the right approach? Did I miss any edge cases? 

> If we have many outgoing requests at a time and some fail, should we consider the connection failed?

I took the approach of "yes". As well as replaying all active requests if a new challenge is received.

I had very successful results using this inside `trin` to facilitate our `utp` transfers, which were really struggling beforehand when concurrent requests were not allowed.

ackintosh:
It has been implemented by #200 and was released in [v0.4.0](https://github.com/sigp/discv5/releases/tag/v0.4.0). 🎉

PR #200, opened 2023-07-10
Add support for concurrent requests to a single peer.

njgheorghita:
## Description

Add support to discv5 for multiple, concurrent network requests to a single peer. See #198 

### How was it fixed?

- Update `ActiveRequests` to support tracking multiple, concurrent, active requests for any given peer.
- Moved `HashMapDelay` to the nonce mapping since it's no longer effective on the peer mapping.
- If any single request fails for a peer, all other active requests for that peer are also failed.
- Replay all active requests if a new challenge is issued.
- Added tests for `ActiveRequests` methods.

## Notes & open questions

- Updated `active_requests` methods to return a `Result` rather than an `Option`. This made sense to me to accurately reflect the cases where the mappings might be out of sync. However, I realize that the `HashMap` & `HashMapDelay` apis return an `Option`, so I don't feel strongly about this change either way.
- This was tested in trin, and resulted in 100% success rate for utp txs over discv5. 

## Change checklist

- [ ] Self-review
- [ ] Tests if relevant

### DIFF
--- a/src/handler/active_requests.rs
+++ b/src/handler/active_requests.rs
@@ -76,9 +76,12 @@ impl ActiveRequests {
         node_address: &NodeAddress,
         id: &RequestId,
     ) -> Option<RequestCall> {
-        match self.active_requests_mapping.entry(node_address.to_owned()) {
+        match self.active_requests_mapping.entry(node_address.clone()) {
             Entry::Vacant(_) => None,
             Entry::Occupied(mut requests) => {
+                if requests.get().is_empty() {
+                    return None;
+                }
                 let index = requests.get().iter().position(|req| {
                     let req_id: RequestId = req.id().into();
                     &req_id == id
@@ -125,6 +128,9 @@ impl Stream for ActiveRequests {
                 match self.active_requests_mapping.entry(node_address.clone()) {
                     Entry::Vacant(_) => Poll::Ready(None),
                     Entry::Occupied(mut requests) => {
+                        if requests.get().is_empty() {
+                            return Poll::Ready(None);
+                        }
                         match requests
                             .get()
                             .iter()
--- a/src/handler/active_requests.rs
+++ b/src/handler/active_requests.rs
@@ -32,6 +32,10 @@ impl ActiveRequests {
             .insert(nonce, node_address);
     }
 
+    pub fn get(&self, node_address: &NodeAddress) -> Option<&Vec<RequestCall>> {
+        self.active_requests_mapping.get(node_address)
+    }
+
     /// Remove a single request identified by its nonce.
     pub fn remove_by_nonce(&mut self, nonce: &MessageNonce) -> Option<(NodeAddress, RequestCall)> {
         let node_address = self.active_requests_nonce_mapping.remove(nonce)?;
--- a/src/handler/active_requests.rs
+++ b/src/handler/active_requests.rs
@@ -87,13 +87,7 @@ impl ActiveRequests {
             .get(node_address)?
             .iter()
             .filter(|req| req.packet().message_nonce() != except)
-            .map(|req| {
-                match req.id() {
-                    HandlerReqId::Internal(id) => id,
-                    HandlerReqId::External(id) => id,
-                }
-                .clone()
-            })
+            .map(|req| req.id().into())
             .collect::<Vec<_>>();
 
         let mut requests = vec![];
--- a/src/handler/active_requests.rs
+++ b/src/handler/active_requests.rs
@@ -1,5 +1,23 @@
 use super::*;
 use delay_map::HashMapDelay;
+use std::fmt;
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ActiveRequestsError {
+    InvalidState,
+}
+
+impl fmt::Display for ActiveRequestsError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            ActiveRequestsError::InvalidState => {
+                write!(f, "Invalid state: active requests mappings are not in sync")
+            }
+        }
+    }
+}
+
+impl std::error::Error for ActiveRequestsError {}
 
 pub(super) struct ActiveRequests {
     /// A list of raw messages we are awaiting a response from the remote.
@@ -34,83 +52,78 @@ impl ActiveRequests {
     }
 
     // Remove a single request identified by its nonce.
-    // Returns None if the nonce is not found.
-    pub fn remove_by_nonce(&mut self, nonce: &MessageNonce) -> Option<(NodeAddress, RequestCall)> {
-        let node_address = match self.active_requests_nonce_mapping.remove(nonce) {
-            Some(node_address) => node_address,
-            None => return None,
-        };
-        let mut requests = match self.active_requests_mapping.remove(&node_address) {
-            Some(requests) => match requests.len() {
-                0 => return None,
-                _ => requests,
-            },
-            None => return None,
-        };
+    pub fn remove_by_nonce(
+        &mut self,
+        nonce: &MessageNonce,
+    ) -> Result<(NodeAddress, RequestCall), ActiveRequestsError> {
+        let node_address = self
+            .active_requests_nonce_mapping
+            .remove(nonce)
+            .ok_or_else(|| ActiveRequestsError::InvalidState)?;
+        let mut requests = self
+            .active_requests_mapping
+            .remove(&node_address)
+            .ok_or_else(|| ActiveRequestsError::InvalidState)?;
         let index = match requests
             .iter()
             .position(|req| req.packet().message_nonce() == nonce)
         {
             Some(index) => index,
             None => {
-                return None;
+                // if nonce req is missing, reinsert remaining requests into mapping
+                self.active_requests_mapping
+                    .insert(node_address.clone(), requests);
+                return Err(ActiveRequestsError::InvalidState);
             }
         };
         let req = requests.remove(index);
         self.active_requests_mapping
             .insert(node_address.clone(), requests);
-        Some((node_address, req))
+        Ok((node_address, req))
     }
 
     // Remove all requests associated with a node.
-    // Returns None if the node is not found.
-    pub fn remove_requests(&mut self, node_address: &NodeAddress) -> Option<Vec<RequestCall>> {
+    pub fn remove_requests(
+        &mut self,
+        node_address: &NodeAddress,
+    ) -> Result<Vec<RequestCall>, ActiveRequestsError> {
         let requests = self
             .active_requests_mapping
             .remove(&node_address)
-            .unwrap_or_default();
+            .ok_or_else(|| ActiveRequestsError::InvalidState)?;
         for req in &requests {
             self.active_requests_nonce_mapping
                 .remove(req.packet().message_nonce());
         }
-        Some(requests)
+        Ok(requests)
     }
 
     // Remove a single request identified by its id.
-    // Returns None if the node is not found.
     pub fn remove_request(
         &mut self,
         node_address: &NodeAddress,
         id: &RequestId,
-    ) -> Option<RequestCall> {
-        let mut reqs = match self.active_requests_mapping.remove(node_address) {
-            Some(reqs) => reqs,
-            None => return None,
-        };
-        match reqs.len() {
-            0 => None,
-            _ => {
-                let index = reqs.iter().position(|req| {
-                    let req_id = match req.id() {
-                        HandlerReqId::Internal(id) | HandlerReqId::External(id) => id,
-                    };
-                    req_id == id
-                });
-                let index = match index {
-                    Some(index) => index,
-                    None => return None,
-                };
-                let req = reqs.remove(index);
-                // Remove the associated nonce mapping.
-                match self
-                    .active_requests_nonce_mapping
-                    .remove(req.packet().message_nonce())
-                {
-                    Some(_) => Some(req),
-                    None => None,
-                }
-            }
-        }
+    ) -> Result<RequestCall, ActiveRequestsError> {
+        let reqs = self
+            .active_requests_mapping
+            .get(node_address)
+            .ok_or_else(|| ActiveRequestsError::InvalidState)?;
+        let index = reqs
+            .iter()
+            .position(|req| {
+                let req_id: RequestId = req.id().into();
+                &req_id == id
+            })
+            .ok_or_else(|| ActiveRequestsError::InvalidState)?;
+        let nonce = reqs
+            .get(index)
+            .ok_or_else(|| ActiveRequestsError::InvalidState)?
+            .packet()
+            .message_nonce()
+            .clone();
+        // Remove the associated nonce mapping.
+        let (_, request_call) = self.remove_by_nonce(&nonce)?;
+        Ok(request_call)
     }
 
     /// Checks that `active_requests_mapping` and `active_requests_nonce_mapping` are in sync.
@@ -127,8 +140,8 @@ impl ActiveRequests {
             }
         }
 
-        for (address, request) in self.active_requests_mapping.iter() {
-            for req in request {
+        for (address, requests) in self.active_requests_mapping.iter() {
+            for req in requests {
                 let nonce = req.packet().message_nonce();
                 if !self.active_requests_nonce_mapping.contains_key(nonce) {
                     panic!("Address {} maps to request with nonce {:?}, which does not exist in `active_requests_nonce_mapping`", address, nonce);
@@ -141,13 +154,16 @@ impl ActiveRequests {
 impl Stream for ActiveRequests {
     type Item = Result<(NodeAddress, RequestCall), String>;
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
-        // should we move timeout management to active_requests_nonce_mapping?
         match self.active_requests_mapping.poll_next_unpin(cx) {
             Poll::Ready(Some(Ok((node_address, mut request_calls)))) => {
                 let request_call = request_calls.remove(0);
+                // reinsert remaining requests into mapping
+                self.active_requests_mapping
+                    .insert(node_address.clone(), request_calls);
+                // remove the nonce mapping
                 self.active_requests_nonce_mapping
                     .remove(request_call.packet().message_nonce())
-                    .expect("fuck");
+                    .expect("Invariant violated: nonce mapping does not exist for request");
                 Poll::Ready(Some(Ok((node_address, request_call))))
             }
             Poll::Ready(Some(Err(err))) => Poll::Ready(Some(Err(err))),
--- a/src/handler/active_requests.rs
+++ b/src/handler/active_requests.rs
@@ -76,6 +76,39 @@ impl ActiveRequests {
         Some(requests)
     }
 
+    /// Remove requests associated with a node, except for the request that has the given message nonce.
+    pub fn remove_requests_except(
+        &mut self,
+        node_address: &NodeAddress,
+        except: &MessageNonce,
+    ) -> Option<Vec<RequestCall>> {
+        let request_ids = self
+            .active_requests_mapping
+            .get(node_address)?
+            .iter()
+            .filter(|req| req.packet().message_nonce() != except)
+            .map(|req| {
+                match req.id() {
+                    HandlerReqId::Internal(id) => id,
+                    HandlerReqId::External(id) => id,
+                }
+                .clone()
+            })
+            .collect::<Vec<_>>();
+
+        let mut requests = vec![];
+        for id in request_ids.iter() {
+            match self.remove_request(node_address, id) {
+                Some(request_call) => requests.push(request_call),
+                None => {
+                    debug_unreachable!("expected to find request with id");
+                    error!("expected to find request with id");
+                }
+            }
+        }
+        Some(requests)
+    }
+
     /// Remove a single request identified by its id.
     pub fn remove_request(
         &mut self,
--- a/src/handler/active_requests.rs
+++ b/src/handler/active_requests.rs
@@ -21,7 +21,7 @@ impl ActiveRequests {
         }
     }
 
-    // Insert a new request into the active requests mapping.
+    /// Insert a new request into the active requests mapping.
     pub fn insert(&mut self, node_address: NodeAddress, request_call: RequestCall) {
         let nonce = *request_call.packet().message_nonce();
         self.active_requests_mapping
@@ -32,37 +32,29 @@ impl ActiveRequests {
             .insert(nonce, node_address);
     }
 
-    // Remove a single request identified by its nonce.
+    /// Remove a single request identified by its nonce.
     pub fn remove_by_nonce(&mut self, nonce: &MessageNonce) -> Option<(NodeAddress, RequestCall)> {
-        let node_address = match self.active_requests_nonce_mapping.remove(nonce) {
-            Some(val) => val,
-            None => return None,
-        };
+        let node_address = self.active_requests_nonce_mapping.remove(nonce)?;
         match self.active_requests_mapping.entry(node_address.clone()) {
             Entry::Vacant(_) => {
                 debug_unreachable!("expected to find node address in active_requests_mapping");
                 error!("expected to find node address in active_requests_mapping");
                 None
             }
-            Entry::Occupied(mut requests) => {
-                let index = requests
-                    .get()
-                    .iter()
-                    .position(|req| req.packet().message_nonce() == nonce)
-                    .expect("to find request call by nonce");
-                Some((node_address, requests.get_mut().remove(index)))
-            }
+            Entry::Occupied(mut requests) => requests
+                .get()
+                .iter()
+                .position(|req| req.packet().message_nonce() == nonce)
+                .map(|index| (node_address, requests.get_mut().remove(index))),
         }
     }
 
-    // Remove all requests associated with a node.
+    /// Remove all requests associated with a node.
     pub fn remove_requests(&mut self, node_address: &NodeAddress) -> Option<Vec<RequestCall>> {
-        let requests = match self.active_requests_mapping.remove(node_address) {
-            Some(val) => val,
-            None => return None,
-        };
+        let requests = self.active_requests_mapping.remove(node_address)?;
         // Account for node addresses in `active_requests_nonce_mapping` with an empty list
         if requests.is_empty() {
+            debug_unreachable!("expected to find requests in active_requests_mapping");
             return None;
         }
         for req in &requests {
@@ -78,7 +70,7 @@ impl ActiveRequests {
         Some(requests)
     }
 
-    // Remove a single request identified by its id.
+    /// Remove a single request identified by its id.
     pub fn remove_request(
         &mut self,
         node_address: &NodeAddress,
@@ -90,13 +82,7 @@ impl ActiveRequests {
                 let index = requests.get().iter().position(|req| {
                     let req_id: RequestId = req.id().into();
                     &req_id == id
-                });
-                let index = match index {
-                    Some(index) => index,
-                    // Node address existence in active requests mapping does not guarantee request
-                    // id existence.
-                    None => return None,
-                };
+                })?;
                 let request_call = requests.get_mut().remove(index);
                 // Remove the associated nonce mapping.
                 self.active_requests_nonce_mapping
@@ -137,14 +123,19 @@ impl Stream for ActiveRequests {
         match self.active_requests_nonce_mapping.poll_next_unpin(cx) {
             Poll::Ready(Some(Ok((nonce, node_address)))) => {
                 match self.active_requests_mapping.entry(node_address.clone()) {
-                    Entry::Vacant(_) => panic!("invalid ActiveRequests state"),
+                    Entry::Vacant(_) => Poll::Ready(None),
                     Entry::Occupied(mut requests) => {
-                        let index = requests
+                        match requests
                             .get()
                             .iter()
                             .position(|req| req.packet().message_nonce() == &nonce)
-                            .expect("to find request call by nonce");
-                        Poll::Ready(Some(Ok((node_address, requests.get_mut().remove(index)))))
+                        {
+                            Some(index) => Poll::Ready(Some(Ok((
+                                node_address,
+                                requests.get_mut().remove(index),
+                            )))),
+                            None => Poll::Ready(None),
+                        }
                     }
                 }
             }
--- a/src/handler/active_requests.rs
+++ b/src/handler/active_requests.rs
@@ -41,11 +41,17 @@ impl ActiveRequests {
                 error!("expected to find node address in active_requests_mapping");
                 None
             }
-            Entry::Occupied(mut requests) => requests
-                .get()
-                .iter()
-                .position(|req| req.packet().message_nonce() == nonce)
-                .map(|index| (node_address, requests.get_mut().remove(index))),
+            Entry::Occupied(mut requests) => {
+                let result = requests
+                    .get()
+                    .iter()
+                    .position(|req| req.packet().message_nonce() == nonce)
+                    .map(|index| (node_address, requests.get_mut().remove(index)));
+                if requests.get().is_empty() {
+                    requests.remove();
+                }
+                result
+            }
         }
     }
 
@@ -79,14 +85,14 @@ impl ActiveRequests {
         match self.active_requests_mapping.entry(node_address.clone()) {
             Entry::Vacant(_) => None,
             Entry::Occupied(mut requests) => {
-                if requests.get().is_empty() {
-                    return None;
-                }
                 let index = requests.get().iter().position(|req| {
                     let req_id: RequestId = req.id().into();
                     &req_id == id
                 })?;
                 let request_call = requests.get_mut().remove(index);
+                if requests.get().is_empty() {
+                    requests.remove();
+                }
                 // Remove the associated nonce mapping.
                 self.active_requests_nonce_mapping
                     .remove(request_call.packet().message_nonce());
@@ -128,18 +134,18 @@ impl Stream for ActiveRequests {
                 match self.active_requests_mapping.entry(node_address.clone()) {
                     Entry::Vacant(_) => Poll::Ready(None),
                     Entry::Occupied(mut requests) => {
-                        if requests.get().is_empty() {
-                            return Poll::Ready(None);
-                        }
                         match requests
                             .get()
                             .iter()
                             .position(|req| req.packet().message_nonce() == &nonce)
                         {
-                            Some(index) => Poll::Ready(Some(Ok((
-                                node_address,
-                                requests.get_mut().remove(index),
-                            )))),
+                            Some(index) => {
+                                let result = (node_address, requests.get_mut().remove(index));
+                                if requests.get().is_empty() {
+                                    requests.remove();
+                                }
+                                Poll::Ready(Some(Ok(result)))
+                            }
                             None => Poll::Ready(None),
                         }
                     }
--- a/src/handler/active_requests.rs
+++ b/src/handler/active_requests.rs
@@ -26,7 +26,7 @@ impl ActiveRequests {
         let nonce = *request_call.packet().message_nonce();
         self.active_requests_mapping
             .entry(node_address.clone())
-            .or_insert_with(Vec::new)
+            .or_default()
             .push(request_call);
         self.active_requests_nonce_mapping
             .insert(nonce, node_address);
@@ -43,10 +43,8 @@ impl ActiveRequests {
                 return;
             };
 
-        self.active_requests_nonce_mapping.insert(
-            new_packet.header.message_nonce.clone(),
-            node_address.clone(),
-        );
+        self.active_requests_nonce_mapping
+            .insert(new_packet.header.message_nonce, node_address.clone());
 
         match self.active_requests_mapping.entry(node_address) {
             Entry::Occupied(mut requests) => {
--- a/src/handler/active_requests.rs
+++ b/src/handler/active_requests.rs
@@ -1,23 +1,7 @@
 use super::*;
 use delay_map::HashMapDelay;
-use std::fmt;
-
-#[derive(Debug, Clone, PartialEq, Eq)]
-pub enum ActiveRequestsError {
-    InvalidState,
-}
-
-impl fmt::Display for ActiveRequestsError {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        match self {
-            ActiveRequestsError::InvalidState => {
-                write!(f, "Invalid state: active requests mappings are not in sync")
-            }
-        }
-    }
-}
-
-impl std::error::Error for ActiveRequestsError {}
+use more_asserts::debug_unreachable;
+use std::collections::hash_map::Entry;
 
 pub(super) struct ActiveRequests {
     /// A list of raw messages we are awaiting a response from the remote.
@@ -40,66 +24,58 @@ impl ActiveRequests {
     // Insert a new request into the active requests mapping.
     pub fn insert(&mut self, node_address: NodeAddress, request_call: RequestCall) {
         let nonce = *request_call.packet().message_nonce();
-        let mut request_calls = self
-            .active_requests_mapping
-            .remove(&node_address)
-            .unwrap_or_default();
-        request_calls.push(request_call);
         self.active_requests_mapping
-            .insert(node_address.clone(), request_calls);
+            .entry(node_address.clone())
+            .or_insert_with(Vec::new)
+            .push(request_call);
         self.active_requests_nonce_mapping
             .insert(nonce, node_address);
     }
 
     // Remove a single request identified by its nonce.
-    pub fn remove_by_nonce(
-        &mut self,
-        nonce: &MessageNonce,
-    ) -> Result<(NodeAddress, RequestCall), ActiveRequestsError> {
-        let node_address = self
-            .active_requests_nonce_mapping
-            .remove(nonce)
-            .ok_or_else(|| ActiveRequestsError::InvalidState)?;
-        let mut requests = self
-            .active_requests_mapping
-            .remove(&node_address)
-            .ok_or_else(|| ActiveRequestsError::InvalidState)?;
-        let index = match requests
-            .iter()
-            .position(|req| req.packet().message_nonce() == nonce)
-        {
-            Some(index) => index,
-            None => {
-                // if nonce req is missing, reinsert remaining requests into mapping
-                if !requests.is_empty() {
-                    self.active_requests_mapping
-                        .insert(node_address.clone(), requests);
-                }
-                return Err(ActiveRequestsError::InvalidState);
-            }
+    pub fn remove_by_nonce(&mut self, nonce: &MessageNonce) -> Option<(NodeAddress, RequestCall)> {
+        let node_address = match self.active_requests_nonce_mapping.remove(nonce) {
+            Some(val) => val,
+            None => return None,
         };
-        let req = requests.remove(index);
-        if !requests.is_empty() {
-            self.active_requests_mapping
-                .insert(node_address.clone(), requests);
+        match self.active_requests_mapping.entry(node_address.clone()) {
+            Entry::Vacant(_) => {
+                debug_unreachable!("expected to find node address in active_requests_mapping");
+                error!("expected to find node address in active_requests_mapping");
+                None
+            }
+            Entry::Occupied(mut requests) => {
+                let index = requests
+                    .get()
+                    .iter()
+                    .position(|req| req.packet().message_nonce() == nonce)
+                    .expect("to find request call by nonce");
+                Some((node_address, requests.get_mut().remove(index)))
+            }
         }
-        Ok((node_address, req))
     }
 
     // Remove all requests associated with a node.
-    pub fn remove_requests(
-        &mut self,
-        node_address: &NodeAddress,
-    ) -> Result<Vec<RequestCall>, ActiveRequestsError> {
-        let requests = self
-            .active_requests_mapping
-            .remove(&node_address)
-            .ok_or_else(|| ActiveRequestsError::InvalidState)?;
-        for req in &requests {
-            self.active_requests_nonce_mapping
-                .remove(req.packet().message_nonce());
+    pub fn remove_requests(&mut self, node_address: &NodeAddress) -> Option<Vec<RequestCall>> {
+        let requests = match self.active_requests_mapping.remove(node_address) {
+            Some(val) => val,
+            None => return None,
+        };
+        // Account for node addresses in `active_requests_nonce_mapping` with an empty list
+        if requests.is_empty() {
+            return None;
         }
-        Ok(requests)
+        for req in &requests {
+            if self
+                .active_requests_nonce_mapping
+                .remove(req.packet().message_nonce())
+                .is_none()
+            {
+                debug_unreachable!("expected to find req with nonce");
+                error!("expected to find req with nonce");
+            }
+        }
+        Some(requests)
     }
 
     // Remove a single request identified by its id.
@@ -107,27 +83,27 @@ impl ActiveRequests {
         &mut self,
         node_address: &NodeAddress,
         id: &RequestId,
-    ) -> Result<RequestCall, ActiveRequestsError> {
-        let reqs = self
-            .active_requests_mapping
-            .get(node_address)
-            .ok_or_else(|| ActiveRequestsError::InvalidState)?;
-        let index = reqs
-            .iter()
-            .position(|req| {
-                let req_id: RequestId = req.id().into();
-                &req_id == id
-            })
-            .ok_or_else(|| ActiveRequestsError::InvalidState)?;
-        let nonce = reqs
-            .get(index)
-            .ok_or_else(|| ActiveRequestsError::InvalidState)?
-            .packet()
-            .message_nonce()
-            .clone();
-        // Remove the associated nonce mapping.
-        let (_, request_call) = self.remove_by_nonce(&nonce)?;
-        Ok(request_call)
+    ) -> Option<RequestCall> {
+        match self.active_requests_mapping.entry(node_address.to_owned()) {
+            Entry::Vacant(_) => None,
+            Entry::Occupied(mut requests) => {
+                let index = requests.get().iter().position(|req| {
+                    let req_id: RequestId = req.id().into();
+                    &req_id == id
+                });
+                let index = match index {
+                    Some(index) => index,
+                    // Node address existence in active requests mapping does not guarantee request
+                    // id existence.
+                    None => return None,
+                };
+                let request_call = requests.get_mut().remove(index);
+                // Remove the associated nonce mapping.
+                self.active_requests_nonce_mapping
+                    .remove(request_call.packet().message_nonce());
+                Some(request_call)
+            }
+        }
     }
 
     /// Checks that `active_requests_mapping` and `active_requests_nonce_mapping` are in sync.
@@ -160,23 +136,17 @@ impl Stream for ActiveRequests {
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
         match self.active_requests_nonce_mapping.poll_next_unpin(cx) {
             Poll::Ready(Some(Ok((nonce, node_address)))) => {
-                // remove the associated mapping
-                let mut reqs = self
-                    .active_requests_mapping
-                    .remove(&node_address)
-                    .ok_or_else(|| ActiveRequestsError::InvalidState)
-                    .unwrap();
-                let index = reqs
-                    .iter()
-                    .position(|req| req.packet().message_nonce() == &nonce)
-                    .ok_or_else(|| ActiveRequestsError::InvalidState)
-                    .unwrap();
-                let req = reqs.remove(index);
-                if reqs.len() > 0 {
-                    self.active_requests_mapping
-                        .insert(node_address.clone(), reqs);
+                match self.active_requests_mapping.entry(node_address.clone()) {
+                    Entry::Vacant(_) => panic!("invalid ActiveRequests state"),
+                    Entry::Occupied(mut requests) => {
+                        let index = requests
+                            .get()
+                            .iter()
+                            .position(|req| req.packet().message_nonce() == &nonce)
+                            .expect("to find request call by nonce");
+                        Poll::Ready(Some(Ok((node_address, requests.get_mut().remove(index)))))
+                    }
                 }
-                Poll::Ready(Some(Ok((node_address, req))))
             }
             Poll::Ready(Some(Err(err))) => Poll::Ready(Some(Err(err))),
             Poll::Ready(None) => Poll::Ready(None),
--- a/src/handler/active_requests.rs
+++ b/src/handler/active_requests.rs
@@ -32,6 +32,42 @@ impl ActiveRequests {
             .insert(nonce, node_address);
     }
 
+    pub fn update_packet(&mut self, old_nonce: MessageNonce, new_packet: Packet) {
+        let node_address =
+            if let Some(node_address) = self.active_requests_nonce_mapping.remove(&old_nonce) {
+                node_address
+            } else {
+                debug_unreachable!("expected to find nonce in active_requests_nonce_mapping");
+                error!("expected to find nonce in active_requests_nonce_mapping");
+                return;
+            };
+
+        self.active_requests_nonce_mapping.insert(
+            new_packet.header.message_nonce.clone(),
+            node_address.clone(),
+        );
+
+        match self.active_requests_mapping.entry(node_address) {
+            Entry::Occupied(mut requests) => {
+                let maybe_request_call = requests
+                    .get_mut()
+                    .iter_mut()
+                    .find(|req| req.packet().message_nonce() == &old_nonce);
+
+                if let Some(request_call) = maybe_request_call {
+                    request_call.update_packet(new_packet);
+                } else {
+                    debug_unreachable!("expected to find request call in active_requests_mapping");
+                    error!("expected to find request call in active_requests_mapping");
+                }
+            }
+            Entry::Vacant(_) => {
+                debug_unreachable!("expected to find node address in active_requests_mapping");
+                error!("expected to find node address in active_requests_mapping");
+            }
+        }
+    }
+
     pub fn get(&self, node_address: &NodeAddress) -> Option<&Vec<RequestCall>> {
         self.active_requests_mapping.get(node_address)
     }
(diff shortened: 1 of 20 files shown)


src/handler/active_requests.rs
```diff
@@ -1,69 +1,133 @@
...
+    // Remove a single request identified by its id.
+    pub fn remove_request(
+        &mut self,
+        node_address: &NodeAddress,
+        id: &RequestId,
+    ) -> Result<RequestCall, ActiveRequestsError> {
+        let reqs = self
+            .active_requests_mapping
+            .get(node_address)
+            .ok_or_else(|| ActiveRequestsError::InvalidState)?;
```
divagant-martian, line 114: Not an invalid internal state

src/handler/active_requests.rs
```diff
@@ -1,69 +1,133 @@
...
+
+    // Remove a single request identified by its id.
+    pub fn remove_request(
+        &mut self,
+        node_address: &NodeAddress,
+        id: &RequestId,
+    ) -> Result<RequestCall, ActiveRequestsError> {
+        let reqs = self
+            .active_requests_mapping
+            .get(node_address)
```
divagant-martian, line 113: should be done with an `entry` to avoid the double access

src/handler/active_requests.rs
```diff
@@ -1,69 +1,133 @@
...
+            })
+            .ok_or_else(|| ActiveRequestsError::InvalidState)?;
+        let nonce = reqs
+            .get(index)
+            .ok_or_else(|| ActiveRequestsError::InvalidState)?
+            .packet()
+            .message_nonce()
+            .clone();
+        // Remove the associated nonce mapping.
+        let (_, request_call) = self.remove_by_nonce(&nonce)?;
```
divagant-martian, line 129: it would be ideal to do this here directly since this involves further checks and accesses already done here

src/handler/active_requests.rs
```diff
@@ -1,69 +1,133 @@
...
+        let index = reqs
+            .iter()
+            .position(|req| {
+                let req_id: RequestId = req.id().into();
+                &req_id == id
+            })
+            .ok_or_else(|| ActiveRequestsError::InvalidState)?;
+        let nonce = reqs
+            .get(index)
+            .ok_or_else(|| ActiveRequestsError::InvalidState)?
```
divagant-martian, line 124: I think this is safe enough to be `expect`d

src/handler/active_requests.rs
```diff
@@ -92,12 +158,25 @@ impl ActiveRequests {
...
+                let index = reqs
+                    .iter()
+                    .position(|req| req.packet().message_nonce() == &nonce)
+                    .ok_or_else(|| ActiveRequestsError::InvalidState)
+                    .unwrap();
+                let req = reqs.remove(index);
+                if reqs.len() > 0 {
+                    self.active_requests_mapping
+                        .insert(node_address.clone(), reqs);
+                }
```
divagant-martian, line 178: do with an `entry`

src/handler/mod.rs
```diff
@@ -187,12 +189,10 @@ pub struct Handler {
     enr: Arc<RwLock<Enr>>,
     /// The key to sign the ENR and set up encrypted communication with peers.
     key: Arc<RwLock<CombinedKey>>,
-    /// Pending raw requests.
+    /// Active requests that are awaiting a response.
     active_requests: ActiveRequests,
     /// The expected responses by SocketAddr which allows packets to pass the underlying filter.
     filter_expected_responses: Arc<RwLock<HashMap<SocketAddr, usize>>>,
-    /// Requests awaiting a handshake completion.
-    pending_requests: HashMap<NodeAddress, Vec<PendingRequest>>,
```
divagant-martian, line 195: Sending concurrent requests should still be done after the handshake, why was this removed?

src/handler/mod.rs
```diff
@@ -891,41 +874,35 @@ impl Handler {
...
+            .active_requests
+            .remove_requests(node_address)
+            .unwrap_or_default();
+        for req in active_requests {
+            let request_id = req.id().clone();
+            if let Err(request_error) = self
+                .send_request::<P>(
+                    req.contact().clone(),
+                    request_id.clone(),
+                    req.body().clone(),
```
divagant-martian, line 888: there should be no need to clone the body and the contact, since the request call as a whole is dropped. It would be better to add a `into_request_parts(self)` (or similar) that gives the three parts that are necessary

src/handler/mod.rs
```diff
@@ -1067,22 +1044,11 @@ impl Handler {
...
-                );
-                // add the request back and reset the timer
-                self.active_requests.insert(node_address, request_call);
-                return;
-            }
-
+        if let Ok(mut request_call) = self
+            .active_requests
+            .remove_request(&node_address, &response.id)
+        {
```
divagant-martian, line 1050: Why remove this logic?

src/handler/active_requests.rs
```diff
@@ -1,69 +1,133 @@
...
     pub fn insert(&mut self, node_address: NodeAddress, request_call: RequestCall) {
         let nonce = *request_call.packet().message_nonce();
+        let mut request_calls = self
+            .active_requests_mapping
+            .remove(&node_address)
+            .unwrap_or_default();
+        request_calls.push(request_call);
         self.active_requests_mapping
-            .insert(node_address.clone(), request_call);
+            .insert(node_address.clone(), request_calls);
```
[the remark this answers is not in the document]
njgheorghita:
```rust
        self.active_requests_mapping
            .entry(node_address.clone())
            .and_modify(|reqs| reqs.push(request_call.clone()))
            .or_insert(vec![request_call]);
```
This is the closest I could get to achieving this with `.entry(..)` and it requires us to implement `Clone` on `RequestCall` which I'm not convinced is justified in this case. Maybe I'm missing a better solution here?

src/handler/active_requests.rs
```diff
@@ -1,69 +1,133 @@
...
+
+impl fmt::Display for ActiveRequestsError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            ActiveRequestsError::InvalidState => {
+                write!(f, "Invalid state: active requests mappings are not in sync")
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
njgheorghita: Decided to go revert to using an `Option` rather than introduce a new `Result`

src/handler/active_requests.rs
```diff
@@ -1,69 +1,133 @@
...
         self.active_requests_mapping
-            .insert(node_address.clone(), request_call);
+            .insert(node_address.clone(), request_calls);
         self.active_requests_nonce_mapping
             .insert(nonce, node_address);
     }
 
-    pub fn get(&self, node_address: &NodeAddress) -> Option<&RequestCall> {
-        self.active_requests_mapping.get(node_address)
-    }
```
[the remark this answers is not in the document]
njgheorghita: Simply because it's no longer used inside `handler/mod.rs`, but happy to re-introduce it.

src/handler/mod.rs
njgheorghita: Ahh, yup makes sense. I re-introduced the `pending_requests` functionality

src/handler/mod.rs
njgheorghita: It's no longer needed, due to the change in `ActiveRequests`. The exception that the removed logic was catching (`response.id` doesn't match request id in `active_requests_mapping` hashmap) is now handled inside `ActiveRequests::remove_request(..)` and will return `None`, which is handled in the `else` statement.

njgheorghita:
Thanks @divagant-martian for the review! I wasn't too familiar with using `entry()` so thanks for pointing out where that would be useful. 
- I went back to the old pattern of returning an `Option` rather than a `Result` from `ActiveRequests::remove_*` methods. Imo it's just cleaner if we want to use `entry()` to mutate `ActiveRequests`. As well as it more closely represents the `HashMap` api. This means that the code functionally ignores scenarios when the `ActiveRequests` mappings are out of sync (besides logging some info). However, this was the behavior before this pr, so I don't consider it a problem.
- I re-introduced `pending_requests` to queue requests that are made before a session is established

src/handler/active_requests.rs
```diff
@@ -1,68 +1,108 @@
...
+            return None;
+        }
+        for req in &requests {
+            if self
+                .active_requests_nonce_mapping
+                .remove(req.packet().message_nonce())
+                .is_none()
+            {
+                debug_unreachable!("expected to find req with nonce");
+                error!("expected to find req with nonce");
```
divagant-martian, line 75: nice

src/handler/active_requests.rs
```diff
@@ -1,68 +1,108 @@
...
+            }
         }
+        Some(requests)
     }
 
-    pub fn remove(&mut self, node_address: &NodeAddress) -> Option<RequestCall> {
-        match self.active_requests_mapping.remove(node_address) {
-            Some(request_call) => {
+    // Remove a single request identified by its id.
```
divagant-martian, line 81:
```suggestion
    /// Remove a single request identified by its id.
```

src/handler/active_requests.rs
```diff
@@ -1,68 +1,108 @@
...
+                let index = requests.get().iter().position(|req| {
+                    let req_id: RequestId = req.id().into();
+                    &req_id == id
+                });
+                let index = match index {
+                    Some(index) => index,
+                    // Node address existence in active requests mapping does not guarantee request
+                    // id existence.
+                    None => return None,
+                };
```
divagant-martian, line 99:
```suggestion
                let index = requests.get().iter().position(|req| {
                    let req_id: RequestId = req.id().into();
                    &req_id == id
                })?;
```

src/handler/active_requests.rs
```diff
@@ -92,12 +134,19 @@ impl ActiveRequests {
...
+        match self.active_requests_nonce_mapping.poll_next_unpin(cx) {
+            Poll::Ready(Some(Ok((nonce, node_address)))) => {
+                match self.active_requests_mapping.entry(node_address.clone()) {
+                    Entry::Vacant(_) => panic!("invalid ActiveRequests state"),
+                    Entry::Occupied(mut requests) => {
+                        let index = requests
+                            .get()
+                            .iter()
+                            .position(|req| req.packet().message_nonce() == &nonce)
+                            .expect("to find request call by nonce");
```
divagant-martian, line 146: better to no panic in those to keep in line with gracefully handling potential inconsistencies

divagant-martian (commented):
partial review for now

ackintosh:
I'm testing this PR with a [simulation](https://github.com/ackintosh/discv5-testground/pull/126). 🙂  The sequence diagram below shows what happens on session expirations.

```mermaid
sequenceDiagram
    participant Node1
    participant Node2
    Note over Node1: Start discv5 server
    Note over Node2: Start discv5 server

    Node1 ->> Node2: message
    Node2 ->> Node1: message

    rect rgb(100, 100, 0)
    Note over Node2: ** Session expired **
    end

    rect rgb(10, 10, 10)
    Note left of Node1: Node1 sends two requests<br> **in parallel**.
    par 
    Node1 ->> Node2: FINDNODE (id:1)
    and 
    Node1 -->> Node2: FINDNODE (id:2)
    end
    end

    Note over Node2: Send WHOAREYOU<br>since the session has been expired
    Node2 ->> Node1: WHOAREYOU (id:1)

    rect rgb(100, 100, 0)
    Note over Node2: Drop FINDNODE(id:2) request<br>since WHOAREYOU already sent.
    end

    Note over Node1: New session established with Node2

    Node1 ->> Node2: Handshake message (id:1)

    Note over Node2: New session established with Node1

    Node2 ->> Node1: NODES (id:1)

    rect rgb(100, 0, 0)
    Note over Node1: ** FINDNODE(id:2) request timed out **
    end
```

I observed that the `FINDNODE (id:2)` request had been dropped and then the request timed out. As fjl said, I think the `FINDNODE (id:2)` request should be replayed with the new session.

https://github.com/sigp/discv5/issues/198#issuecomment-1616678592
> Remember that in discv5, it is assumed the session may be dropped at any moment. If the node has multiple requests in flight with a peer, and the peer answers with a challenge (WHOAREYOU), all pending requests would need to be replayed.

src/handler/mod.rs
```diff
@@ -494,14 +501,23 @@ impl Handler {
...
-                // We are initiating a new session
+                // Queue the request for sending after the handshake completes
+                self.pending_requests
+                    .entry(node_address.clone())
+                    .or_insert_with(Vec::new)
+                    .push(PendingRequest {
+                        contact: contact.clone(),
+                        request_id: request_id.clone(),
+                        request: request.clone(),
+                    });
```
ackintosh, line 519: I think this is unnecessary. The entry in `pending_requests` inserted here remains despite the request having been completed.

```mermaid

sequenceDiagram
    participant Node1
    participant Node2

    Note over Node1: Send FINDNODE request to Node2.

    rect rgb(10, 10, 10)
    Note left of Node1: Handler::send_request()
    Note over Node1: Send random packet <br> since no session exists.
    Note over Node1: pending_requests.insert
    Note over Node1: active_requests.insert
    end



    Node1 ->> Node2: Random packet
    Node2 ->> Node1: WHOAREYOU

    rect rgb(10, 10, 10)
    Note left of Node1: Handler::handle_challenge()
    Note over Node1: active_requests.remove_by_nonce
    Note over Node1: Generate a new session <br> and create authentication packet
    Note over Node1: active_requests.insert
    end

    Node1 ->> Node2: Handshake message 
    Node2 ->> Node1: NODES

    rect rgb(10, 10, 10)
    Note left of Node1: Handler::handle_response()
    Note over Node1: active_requests.remove_request
    end

    rect rgb(100, 0, 0)
    Note over Node1: The request has completed <br> but the entry in pending_requests remains.
    end

```

src/handler/mod.rs
```diff
@@ -1241,6 +1261,27 @@ impl Handler {
...
+                    if let Err(e) = self
+                        .service_send
+                        .send(HandlerOut::RequestFailed(id.clone(), error.clone()))
+                        .await
+                    {
+                        warn!("Failed to inform request failure {e}")
+                    }
+                }
+            }
+        }
```
ackintosh, line 1284: We should call `self.remove_expected_response()` here to align the number of expected responses.

njgheorghita:
@ackintosh Thanks for the review! Made the suggested changes. Now, all active requests are replayed if an active session is re-established. Tested it out locally with testground and it appears to have solved the `FINDNODE` request timeout. Lmk if there are any other changes / testground simulations to run. 

Btw, curious as to how you're finding testground? Looks like a pretty useful tool and something that we might want to use in trin.

ackintosh:
@njgheorghita Thanks! I will review the changes.

Yeah testground is so helpful. It's easy to simulate/reproduce particular situation/problem with testground. We also using testground for the gossipsub protocol https://github.com/sigp/gossipsub-testground. I hope this can be of reference.

src/handler/mod.rs
ackintosh:
```suggestion
        // fail all active requests
        for req in self
            .active_requests
            .remove_requests(node_address)
            .unwrap_or_default()
        {
            match req.id() {
                HandlerReqId::Internal(_) => {
                    // Do not report failures on requests belonging to the handler.
                }
                HandlerReqId::External(id) => {
                    if let Err(e) = self
                        .service_send
                        .send(HandlerOut::RequestFailed(id.clone(), error.clone()))
                        .await
                    {
                        warn!("Failed to inform request failure {e}")
                    }
                    self.remove_expected_response(node_address.socket_addr);
                }
            }
        }
```

src/handler/mod.rs
```diff
@@ -1241,6 +1256,28 @@ impl Handler {
...
+        self.remove_expected_response(node_address.socket_addr);
```
ackintosh, line 1280: Sorry my comment was wrong. 😅 `self.remove_expected_response` should be here because the same number of active_requests removed must be subtracted from the expected responses.

```suggestion
        for req in self
            .active_requests
            .remove_requests(node_address)
            .unwrap_or_default()
        {
            match req.id() {
                HandlerReqId::Internal(_) => {
                    // Do not report failures on requests belonging to the handler.
                }
                HandlerReqId::External(id) => {
                    if let Err(e) = self
                        .service_send
                        .send(HandlerOut::RequestFailed(id.clone(), error.clone()))
                        .await
                    {
                        warn!("Failed to inform request failure {e}")
                    }
                }
            }
            self.remove_expected_response(node_address.socket_addr);
        }
```

src/handler/mod.rs
```diff
@@ -1152,9 +1161,14 @@ impl Handler {
...
 
-    fn new_session(&mut self, node_address: NodeAddress, session: Session) {
+    async fn new_session<P: ProtocolIdentity>(
+        &mut self,
+        node_address: NodeAddress,
+        session: Session,
+    ) {
         if let Some(current_session) = self.sessions.get_mut(&node_address) {
             current_session.update(session);
+            self.replay_active_requests::<P>(&node_address).await;
```
divagant-martian, line 1171: changes like this deserve a comment.
I need to think this one a bit more but the session stores the old keys in case they are needed, so I'm not sure this is necessary

src/handler/mod.rs
```diff
@@ -1210,7 +1224,7 @@ impl Handler {
             .await;
     }
 
-    /// Removes a session and updates associated metrics and fields.
+    /// Removes a session, fails all of that session's active & pending requests, and updates associated metrics and fields.
```
divagant-martian, line 1227: pending requests can't be failed since they haven't been tried

src/handler/mod.rs
divagant-martian, line 1280: the count of expected responses depends on the number of sent requests, so this needs to be moved up

divagant-martian (changes requested):
This is progressing in the right direction but there are some bugs to be fixed

src/handler/active_requests.rs
```diff
@@ -92,12 +120,24 @@ impl ActiveRequests {
...
+                            .get()
+                            .iter()
+                            .position(|req| req.packet().message_nonce() == &nonce)
+                        {
+                            Some(index) => Poll::Ready(Some(Ok((
+                                node_address,
+                                requests.get_mut().remove(index),
+                            )))),
+                            None => Poll::Ready(None),
+                        }
```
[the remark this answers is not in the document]
njgheorghita: > this will leave requests with empty lists

Not sure I'm understanding your suggestion here. Do you want to see an empty requests replaced with `Entry::Vacant` after removing the last request? Based on your suggestion on line 89, it seems like the convention is that it is possible for the `active_requests_mapping` entry for a node to be an empty list instead of strictly `Entry::Vacant`.

In the case where `requests` is an empty list, this logic will simply return `None => Poll::Ready(None)`. There doesn't seem to be a need to handle the empty list case any differently? But I could be missing something.

But if you'd like to see the change where the `active_requests_mapping` entry for a node without any active requests **cannot** be an empty list and **must** always be `Entry::Vacant`, I can make those changes.

src/handler/mod.rs
njgheorghita: Added a comment, but lmk if you feel like this needs to be removed. Although just to mention that this was the change that solved the `FindNode` requests timeout error described above, but if this is unnecessary I'll debug that failure some more and find a more appropriate solution

src/handler/active_requests.rs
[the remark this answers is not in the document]
divagant-martian: To be clear, on empty list use `entry.remove()`

src/handler/mod.rs
ackintosh: [The timeout](https://github.com/sigp/discv5/pull/200#issuecomment-1657275126) has been solved by this change however I noticed another issue about this replaying:

As highlighted by the red square below, `replay_active_requests()` re-sends the `FINDNODE(id:1)` that has already been sent in the Handshake message. Active requests that should be re-sent is only `FINDNODE(id:2)` in the case of this example.

```mermaid
sequenceDiagram
    participant Node1
    participant Node2
    Note over Node1: Start discv5 server
    Note over Node2: Start discv5 server

    Node1 ->> Node2: message
    Node2 ->> Node1: message

    rect rgb(100, 100, 0)
    Note over Node2: ** Session expired **
    end

    rect rgb(10, 10, 10)
    Note left of Node1: Node1 sends two requests<br> **in parallel**.
    par 
    Node1 ->> Node2: FINDNODE (id:1)
    and 
    Node1 -->> Node2: FINDNODE (id:2)
    end
    end

    Note over Node2: Send WHOAREYOU<br>since the session has been expired
    Node2 ->> Node1: WHOAREYOU (id:1)

    rect rgb(100, 100, 0)
    Note over Node2: Drop FINDNODE(id:2) request<br>since WHOAREYOU already sent.
    end

    Note over Node1: New session established with Node2

    Node1 ->> Node2: Handshake message (id:1)

    Note over Node2: New session established with Node1

    rect rgb(10, 10, 10)
    Note left of Node1: Handler::replay_active_requests()
    rect rgb(100, 0, 0)
    Node1 ->> Node2: FINDNODE (id:1) <br>**duplicates with Handshake message (id:1)**
    end
    Node1 -->> Node2: FINDNODE (id:2)
    end

    Node2 ->> Node1: NODES (id:1)
    Node2 -->> Node1: NODES (id:2)

    rect rgb(100, 0, 0)
    Node2 ->> Node1: NODES (id:1)
    end
```

src/handler/mod.rs
```diff
@@ -331,13 +340,13 @@ impl Handler {
...
+                Some(Ok((node_address, active_request))) = self.active_requests.next() => {
+                    self.handle_request_timeout(node_address, active_request).await;
                 }
                 Some(Ok((node_address, _challenge))) = self.active_challenges.next() => {
-                    // A challenge has expired. There could be pending requests awaiting this
-                    // challenge. We process them here
-                    self.send_next_request::<P>(node_address).await;
+                    // A challenge has expired. There could be active requests impacted by this
+                    // challenge. We replay them here
+                    self.replay_active_requests::<P>(&node_address).await;
```
ackintosh, line 349: I think we need to send pending_requests here not active_requests.

Current implementation, requests in pending_requests remain without being sent in case of WHOAREYOU timeout:

```mermaid
sequenceDiagram
    participant Node1
    participant Node2


    Node2 ->> Node1: Random packet
    Note over Node1: active_challenges.insert()
    Node1 ->> Node2: WHOAREYOU
    rect rgb(100, 100, 0)
    Note over Node2: ** Discard the WHOAREYOU packet<br>to reproduce kind of network issue. **
    end

    rect rgb(10, 10, 10)
    Note left of Node1: Node1 want to send FINDNODE to Node2<br>but active_challenge exists.<br>So insert requests into pending_requests.
    par 
    Note over Node1: pending_requests.insert()
    and 
    Note over Node1: pending_requests.insert()
    end
    end

    Note over Node1: The challenge in active_challenges<br>has been expired.

    rect rgb(100, 0, 0)
    Note over Node1: FINDNODE requests remain in pending_requests without being sent.
    end
```

src/handler/mod.rs
njgheorghita: Cool, thanks for the diagram it was super helpful. I believe I've implemented a sufficient solution that filters out the handshake request when replaying active requests. But only in the case when `Node2` sends the `whoareyou`. I'm not sure if I took the right approach here, but lmk if you see anything that needs changing.

njgheorghita:
Just a note that I've been experiencing some flaky behavior while trying to use this branch inside trin. Looks like there could be a problem with how requests are queued for a node before the session is established. Will update here once I've found a solution.

ackintosh:
@njgheorghita Could you please merge current master into this branch? There're some fixes. I'm not sure if the fixes relates to that problem though.

njgheorghita:
@ackintosh Doesn't seem to have fixed the flakiness, I should have some time in the next day or two to further investigate what's going on there and I'll report back here

ackintosh:
@njgheorghita I'm working on adding a test for `replay_active_request()` and trying to fix bugs I found through the test.

I don't have permission to push commits to this PR directly, so I have created a PR to your `experimental` branch here: https://github.com/njgheorghita/discv5/pull/2

njgheorghita:
@ackintosh Any familiarity with the failing test? It doesn't seem to me like it's a result of this pr, so maybe it's a previously-known flaky test?

Anyways, still experiencing flakiness using this pr in our trin testing suite. As far as I can tell, the flakiness is a result of this pr and not trin's application-layer code. I'd feel more comfortable identifying the cause of the flakiness before merging this through. From what I can tell, there's just some mis-management of pending & active requests going on. When I write unit-tests to expose the issue, everything works fine. However, in trin, since we implement our sub-protocols over the base-layer `TalkReq`, we tend to hammer out quite a lot of successive `TalkReq` messages, and it appears like some of those are dropped while we are establishing the session. 

I'm off this next week, but resolving this will be my top priority when I return the following week.

ackintosh:
@njgheorghita 

> Any familiarity with the failing test? It doesn't seem to me like it's a result of this pr, so maybe it's a previously-known flaky test?

Oh, that's my mistake while adding the test. 😓 I have created a PR to fix the flakiness. Please have a look when you have time.

https://github.com/njgheorghita/discv5/pull/3

> From what I can tell, there's just some mis-management of pending & active requests going on.

Interesting... I will also investigate that.

ackintosh:
A problem I noticed is that if a node sends multiple requests in parallel **before establishing a session** with the destination node, some requests are dropped by the destination node and those requests timed out.

@njgheorghita Seems to relate to the problem happening in trin?

```mermaid
sequenceDiagram
    participant Node1
    participant Node2

    Note over Node1: No session with Node2

    rect rgb(10, 10, 10)
    Note left of Node1: Node1 attempts to send requests in parallel<br> but no session with Node2.<br> So Node1 sends random packets in parallel.
    par 
    Node1 ->> Node2: Random packet (id:1)
    Note over Node1: Insert the request into `active_requests`
    and 
    Node1 ->> Node2: Random packet (id:2)
    Note over Node1: Insert the request into `active_requests`
    and 
    Node1 ->> Node2: Random packet (id:3)
    Note over Node1: Insert the request into `active_requests`
    end
    end

    Node2 ->> Node1: WHOAREYOU (id:1)
    rect rgb(100, 100, 0)
    Note over Node2: Drop the requests (id:2, 3)<br>since WHOAREYOU already sent.
    end

    Note over Node1: New session established with Node2
    Node1 ->> Node2: Handshake message (id:1)

    Note over Node2: New session established with Node1
    Node2 ->> Node1: Response (id:1)

    rect rgb(0, 100, 0)
    Note over Node1: The request (id:1) completed.
    end

    rect rgb(100, 0, 0)
    Note over Node1: The request (id:2) timed out.
    Note over Node1: The request (id:3) timed out.
    end
```

ackintosh:
To fix the timeout issue I've reported in https://github.com/sigp/discv5/pull/200#issuecomment-1710867435, I believe that storing requests into `pending_requests` and send the pending requests once a session is established will help.

I'm working on adding tests and fixes for this issue.

```mermaid
sequenceDiagram
    participant Node1
    participant Node2

    Note over Node1: No session with Node2

    rect rgb(10, 10, 10)
    Note left of Node1: Node1 attempts to send multiple requests in parallel <br> but no session with Node2.<br> So Node1 sends a random packet for the first request, <br>and the rest of the requests are inserted into pending_requests.
    par 
    Node1 ->> Node2: Random packet (id:1)
    Note over Node1: Insert the request into `active_requests`
    and 
    Note over Node1: Insert Request(id:2) into *pending_requests*
    and 
    Note over Node1: Insert Request(id:3) into *pending_requests*
    end
    end

    Node2 ->> Node1: WHOAREYOU (id:1)

    Note over Node1: New session established with Node2
    Node1 ->> Node2: Handshake message (id:1)

    Note over Node2: New session established with Node1

    rect rgb(0, 100, 0)
    Note over Node1: Send pending requests since a session has been established.
    Node1 ->> Node2: Request (id:2)
    Node1 ->> Node2: Request (id:3)
    end

    Node2 ->> Node1: Response (id:1)
    Node2 ->> Node1: Response (id:2)
    Node2 ->> Node1: Response (id:3)

    Note over Node1: The request (id:1) completed.
    Note over Node1: The request (id:2) completed.
    Note over Node1: The request (id:3) completed.
```

ackintosh:
@njgheorghita I have created https://github.com/njgheorghita/discv5/pull/4 to fix the timeout issue. Please have a look when you have time. 🙏

ackintosh:
We have tested and fixed bugs on the Handler so far. Now I'm checking on the Service.

ackintosh:
I found a bug in the Service module. Since [active_nodes_responses](https://github.com/sigp/discv5/blob/d2e30e04ee62418b9e57278cee907c02b99d5bd1/src/service.rs#L189-L190) is defined as `HashMap<NodeId, NodesResponse>`, Service can not handle concurrent NODES responses from a single node correctly. This bug happens when NODES response is sent as multiple messages.

Here is a sequence diagram to show how the bug happens:

```mermaid
sequenceDiagram
    participant Node1
    participant Node2

    rect rgb(10, 10, 10)
    Note left of Node1: Node1 sends FINDNODE requests in parallel.
    par 
    Node1 ->> Node2: Message(FINDNODE id:1)
    and
    Node1 ->> Node2: Message(FINDNODE id:2)
    end
    end

    rect rgb(10, 10, 10)
    Note right of Node2: The responses to FINDNODE(id:1) are sent as multiple messages. <br> NODES(id:1) is `total: 2`.
    Node2 -->> Node1: NODES (id:1) (1 of 2)
    Note over Node1: Service module stores `NODES(id:1) (1 of 2)` <br>in `self.active_nodes_responses` <br>and waits for the second response.
    end

    rect rgb(10, 10, 10)
    Note left of Node1: In this test case, <br>we suppose `NODES(id:2)` reaches Node1 <br>before `NODES(id:1) *(2 of 2)*`, <br>due to reasons such as network latency.
    Node2 -->> Node1: NODES (id:2) (1 of 1)
    end

    rect rgb(100, 0, 0)
    Note over Node1: In Service module, <br>`NODES (id:1) (1 of 2)` is dropped here <br>because Service can not handle concurrent responses.
    end

    Node2 -->> Node1: NODES (id:1) (2 of 2)
```

ackintosh:
I'm working on the bug in the Service module.

ackintosh:
@njgheorghita I have filed https://github.com/njgheorghita/discv5/pull/5 to fix [this bug](https://github.com/sigp/discv5/pull/200#issuecomment-1732134151) to your `experimental` branch. Please have a look when you have time. 🙏

ackintosh:
@divagant-martian @AgeManning My review of this PR is complete. I would appreciate it if you could have a look at this.

AgeManning (commented):
This looks pretty good. 

I honestly thought it was going to be harder dealing with the session logic and various pitfalls that can happen with message re-ordering/message loss and session establishment. 

Has anyone run this on a live network and verified that it works as expected?

KolbyML:
We are currently using it on Portal's testnet which is more like a mainnet. And it has increased throughput a lot and performance alot.

carver:
To be clear, we have a handful of other failures on our network that are unrelated (to the best of our knowledge). They are obscuring results, so it's a little difficult to reliably confirm that we aren't seeing *any* problems related to this change. But nothing specific that we've identified yet! (and it's clearly an improvement for us, because we send a bunch of parallel `TALKREQ` messages)

src/handler/mod.rs
```diff
@@ -813,17 +832,18 @@ impl Handler {
...
+                        self.handle_message(
                             node_address.clone(),
                             message_nonce,
                             message,
                             authenticated_data,
                         )
                         .await;
-                        // We could have pending messages that were awaiting this session to be
-                        // established. If so process them.
-                        self.send_next_request::<P>(node_address).await;
```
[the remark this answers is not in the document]
njgheorghita: Yup, those are now handled inside `new_session()` when a new session is established.

ackintosh:
I'm testing this PR by updating lighthouse and running boot node, and experiencing weird behavior that the boot node repeats establishing sessions to peers. 🤔 

```bash
$ lighthouse-update-discv5 --version   
Lighthouse v4.5.0-441fc16+
BLS library: blst
SHA256 hardware acceleration: false
Allocator: jemalloc
Profile: release
Specs: mainnet (true), minimal (false), gnosis (false)

$ lighthouse-update-discv5 boot_node --enr-address 133.32.132.191 --debug-level trace
...
...
Oct 02 22:32:28.438 DEBG Sending RPC Request: id: 912e3b11fb4ec960: FINDNODE Request: distance: [247, 248, 246] to node: Node: 0x8114..1f57, addr: 52.16.75.63:39999
...
Oct 02 22:32:28.438 TRCE Starting session. Sending random packet to: Node: 0x8114..1f57, addr: 52.16.75.63:39999
...
Oct 02 22:32:28.728 TRCE Received a WHOAREYOU packet response. Source: Node: 0x8114..1f57, addr: 52.16.75.63:39999
Oct 02 22:32:28.728 TRCE Sending Authentication response to node: Node: 0x8114..1f57, addr: 52.16.75.63:39999 (External(RequestId([145, 46, 59, 17, 251, 78, 201, 96])))
Oct 02 22:32:28.728 DEBG Session established with Node: 0x8114..1f57, direction: Outgoing
Oct 02 22:32:28.728 DEBG New connected node added to routing table: 0x8114..1f57
Oct 02 22:32:28.728 DEBG Sending RPC Request: id: 86ebef626a7bd121: PING: enr_seq: 448 to node: Node: 0x8114..1f57, addr: 52.16.75.63:39999
...
...
Oct 02 22:32:29.007 TRCE Received a WHOAREYOU packet response. Source: Node: 0x8114..1f57, addr: 52.16.75.63:39999
Oct 02 22:32:29.008 TRCE Sending Authentication response to node: Node: 0x8114..1f57, addr: 52.16.75.63:39999 (External(RequestId([134, 235, 239, 98, 106, 123, 209, 33])))
Oct 02 22:32:29.008 TRCE Replaying active requests. Node: 0x8114..1f57, addr: 52.16.75.63:39999, Some([185, 182, 195, 79, 25, 234, 155, 157, 160, 60, 73, 168])
Oct 02 22:32:29.008 TRCE Active request to be replayed. 912e3b11fb4ec960, Node: 0x8114..1f57, addr: 52.16.75.63:39999, FINDNODE Request: distance: [247, 248, 246]
Oct 02 22:32:29.008 DEBG Session established with Node: 0x8114..1f57, direction: Outgoing
...
...
Oct 02 22:32:29.293 TRCE Received a WHOAREYOU packet response. Source: Node: 0x8114..1f57, addr: 52.16.75.63:39999
Oct 02 22:32:29.293 TRCE Sending Authentication response to node: Node: 0x8114..1f57, addr: 52.16.75.63:39999 (External(RequestId([145, 46, 59, 17, 251, 78, 201, 96])))
Oct 02 22:32:29.293 TRCE Replaying active requests. Node: 0x8114..1f57, addr: 52.16.75.63:39999, Some([74, 159, 87, 43, 83, 128, 61, 156, 222, 153, 65, 116])
Oct 02 22:32:29.293 TRCE Active request to be replayed. 86ebef626a7bd121, Node: 0x8114..1f57, addr: 52.16.75.63:39999, PING: enr_seq: 448
Oct 02 22:32:29.293 DEBG Session established with Node: 0x8114..1f57, direction: Outgoing
...
...
Oct 02 22:32:29.571 TRCE Received a WHOAREYOU packet response. Source: Node: 0x8114..1f57, addr: 52.16.75.63:39999
Oct 02 22:32:29.571 TRCE Sending Authentication response to node: Node: 0x8114..1f57, addr: 52.16.75.63:39999 (External(RequestId([134, 235, 239, 98, 106, 123, 209, 33])))
Oct 02 22:32:29.571 TRCE Replaying active requests. Node: 0x8114..1f57, addr: 52.16.75.63:39999, Some([4, 53, 1, 184, 247, 199, 235, 195, 150, 213, 47, 184])
Oct 02 22:32:29.571 TRCE Active request to be replayed. 912e3b11fb4ec960, Node: 0x8114..1f57, addr: 52.16.75.63:39999, FINDNODE Request: distance: [247, 248, 246]
Oct 02 22:32:29.571 DEBG Session established with Node: 0x8114..1f57, direction: Outgoing
...
...

**Comment by AgeManning** [comment]
Do you get responses from that node from your request @ackintosh ?

**Comment by ackintosh** [comment]
No I don't get responses from that node. Here is other logs grepped by `0x1645..b7ea`. Seems the node doesn't return responses but WHOAREYOU only. Still I'm digging into this, probably we are not handling this case propery.

<details><summary>Logs</summary>
<p>

```
Oct 03 22:36:53.312 DEBG Sending RPC Request: id: 3b870c12b1f265b2: FINDNODE Request: distance: [245, 246, 244] to node: Node: 0x1645..b7ea, addr: 58.96.74.212:9001
Oct 03 22:36:53.312 TRCE Starting session. Sending random packet to: Node: 0x1645..b7ea, addr: 58.96.74.212:9001
Oct 03 22:36:53.486 TRCE Received a WHOAREYOU packet response. Source: Node: 0x1645..b7ea, addr: 58.96.74.212:9001
Oct 03 22:36:53.486 TRCE Sending Authentication response to node: Node: 0x1645..b7ea, addr: 58.96.74.212:9001 (External(RequestId([59, 135, 12, 18, 177, 242, 101, 178])))
Oct 03 22:36:53.486 DEBG Session established with Node: 0x1645..b7ea, direction: Outgoing
Oct 03 22:36:53.486 DEBG New connected node added to routing table: 0x1645..b7ea
Oct 03 22:36:53.486 DEBG Sending RPC Request: id: 0c9ea7b221a03aff: PING: enr_seq: 448 to node: Node: 0x1645..b7ea, addr: 58.96.74.212:9001
Oct 03 22:36:53.661 TRCE Received a WHOAREYOU packet response. Source: Node: 0x1645..b7ea, addr: 58.96.74.212:9001
Oct 03 22:36:53.662 TRCE Sending Authentication response to node: Node: 0x1645..b7ea, addr: 58.96.74.212:9001 (External(RequestId([12, 158, 167, 178, 33, 160, 58, 255])))
Oct 03 22:36:53.662 TRCE Replaying active requests. Node: 0x1645..b7ea, addr: 58.96.74.212:9001, Some([27, 229, 112, 152, 209, 130, 102, 46, 27, 179, 100, 208])
Oct 03 22:36:53.662 TRCE Active request to be replayed. 3b870c12b1f265b2, Node: 0x1645..b7ea, addr: 58.96.74.212:9001, FINDNODE Request: distance: [245, 246, 244]
Oct 03 22:36:53.662 DEBG Session established with Node: 0x1645..b7ea, direction: Outgoing
Oct 03 22:36:53.838 TRCE Received a WHOAREYOU packet response. Source: Node: 0x1645..b7ea, addr: 58.96.74.212:9001
Oct 03 22:36:53.839 TRCE Sending Authentication response to node: Node: 0x1645..b7ea, addr: 58.96.74.212:9001 (External(RequestId([59, 135, 12, 18, 177, 242, 101, 178])))
Oct 03 22:36:53.839 TRCE Replaying active requests. Node: 0x1645..b7ea, addr: 58.96.74.212:9001, Some([185, 198, 255, 223, 188, 195, 30, 244, 87, 255, 88, 47])
Oct 03 22:36:53.839 TRCE Active request to be replayed. 0c9ea7b221a03aff, Node: 0x1645..b7ea, addr: 58.96.74.212:9001, PING: enr_seq: 448
Oct 03 22:36:53.839 DEBG Session established with Node: 0x1645..b7ea, direction: Outgoing
Oct 03 22:36:54.013 TRCE Received a WHOAREYOU packet response. Source: Node: 0x1645..b7ea, addr: 58.96.74.212:9001
Oct 03 22:36:54.013 TRCE Sending Authentication response to node: Node: 0x1645..b7ea, addr: 58.96.74.212:9001 (External(RequestId([12, 158, 167, 178, 33, 160, 58, 255])))
Oct 03 22:36:54.013 TRCE Replaying active requests. Node: 0x1645..b7ea, addr: 58.96.74.212:9001, Some([67, 201, 43, 220, 127, 62, 21, 108, 44, 120, 186, 46])
Oct 03 22:36:54.013 TRCE Active request to be replayed. 3b870c12b1f265b2, Node: 0x1645..b7ea, addr: 58.96.74.212:9001, FINDNODE Request: distance: [245, 246, 244]
Oct 03 22:36:54.014 DEBG Session established with Node: 0x1645..b7ea, direction: Outgoing
Oct 03 22:36:54.191 TRCE Received a WHOAREYOU packet response. Source: Node: 0x1645..b7ea, addr: 58.96.74.212:9001
Oct 03 22:36:54.192 TRCE Sending Authentication response to node: Node: 0x1645..b7ea, addr: 58.96.74.212:9001 (External(RequestId([59, 135, 12, 18, 177, 242, 101, 178])))
Oct 03 22:36:54.192 TRCE Replaying active requests. Node: 0x1645..b7ea, addr: 58.96.74.212:9001, Some([195, 6, 58, 33, 142, 100, 251, 199, 164, 48, 95, 103])
Oct 03 22:36:54.192 TRCE Active request to be replayed. 0c9ea7b221a03aff, Node: 0x1645..b7ea, addr: 58.96.74.212:9001, PING: enr_seq: 448
Oct 03 22:36:54.192 DEBG Session established with Node: 0x1645..b7ea, direction: Outgoing
Oct 03 22:36:54.375 TRCE Received a WHOAREYOU packet response. Source: Node: 0x1645..b7ea, addr: 58.96.74.212:9001
Oct 03 22:36:54.375 TRCE Sending Authentication response to node: Node: 0x1645..b7ea, addr: 58.96.74.212:9001 (External(RequestId([12, 158, 167, 178, 33, 160, 58, 255])))
Oct 03 22:36:54.375 TRCE Replaying active requests. Node: 0x1645..b7ea, addr: 58.96.74.212:9001, Some([132, 14, 33, 133, 180, 28, 203, 184, 229, 75, 36, 120])
Oct 03 22:36:54.376 TRCE Active request to be replayed. 3b870c12b1f265b2, Node: 0x1645..b7ea, addr: 58.96.74.212:9001, FINDNODE Request: distance: [245, 246, 244]
Oct 03 22:36:54.376 DEBG Session established with Node: 0x1645..b7ea, direction: Outgoing
Oct 03 22:36:54.550 TRCE Received a WHOAREYOU packet response. Source: Node: 0x1645..b7ea, addr: 58.96.74.212:9001
Oct 03 22:36:54.550 TRCE Sending Authentication response to node: Node: 0x1645..b7ea, addr: 58.96.74.212:9001 (External(RequestId([59, 135, 12, 18, 177, 242, 101, 178])))
Oct 03 22:36:54.550 TRCE Replaying active requests. Node: 0x1645..b7ea, addr: 58.96.74.212:9001, Some([132, 56, 8, 137, 142, 80, 22, 203, 106, 57, 21, 169])
Oct 03 22:36:54.550 TRCE Active request to be replayed. 0c9ea7b221a03aff, Node: 0x1645..b7ea, addr: 58.96.74.212:9001, PING: enr_seq: 448
Oct 03 22:36:54.550 DEBG Session established with Node: 0x1645..b7ea, direction: Outgoing
Oct 03 22:36:54.724 TRCE Received a WHOAREYOU packet response. Source: Node: 0x1645..b7ea, addr: 58.96.74.212:9001
Oct 03 22:36:54.724 TRCE Sending Authentication response to node: Node: 0x1645..b7ea, addr: 58.96.74.212:9001 (External(RequestId([12, 158, 167, 178, 33, 160, 58, 255])))
Oct 03 22:36:54.724 TRCE Replaying active requests. Node: 0x1645..b7ea, addr: 58.96.74.212:9001, Some([155, 72, 208, 213, 96, 218, 203, 252, 118, 83, 187, 163])
Oct 03 22:36:54.724 TRCE Active request to be replayed. 3b870c12b1f265b2, Node: 0x1645..b7ea, addr: 58.96.74.212:9001, FINDNODE Request: distance: [245, 246, 244]
Oct 03 22:36:54.724 DEBG Session established with Node: 0x1645..b7ea, direction: Outgoing
Oct 03 22:36:54.898 TRCE Received a WHOAREYOU packet response. Source: Node: 0x1645..b7ea, addr: 58.96.74.212:9001
Oct 03 22:36:54.898 TRCE Sending Authentication response to node: Node: 0x1645..b7ea, addr: 58.96.74.212:9001 (External(RequestId([59, 135, 12, 18, 177, 242, 101, 178])))
Oct 03 22:36:54.898 TRCE Replaying active requests. Node: 0x1645..b7ea, addr: 58.96.74.212:9001, Some([187, 121, 39, 52, 62, 161, 235, 15, 47, 192, 180, 23])
Oct 03 22:36:54.898 TRCE Active request to be replayed. 0c9ea7b221a03aff, Node: 0x1645..b7ea, addr: 58.96.74.212:9001, PING: enr_seq: 448
Oct 03 22:36:54.901 DEBG Session established with Node: 0x1645..b7ea, direction: Outgoing
Oct 03 22:36:55.071 TRCE Received a WHOAREYOU packet response. Source: Node: 0x1645..b7ea, addr: 58.96.74.212:9001
Oct 03 22:36:55.072 TRCE Sending Authentication response to node: Node: 0x1645..b7ea, addr: 58.96.74.212:9001 (External(RequestId([12, 158, 167, 178, 33, 160, 58, 255])))
Oct 03 22:36:55.072 TRCE Replaying active requests. Node: 0x1645..b7ea, addr: 58.96.74.212:9001, Some([45, 48, 62, 49, 46, 185, 158, 42, 171, 123, 37, 79])
Oct 03 22:36:55.072 TRCE Active request to be replayed. 3b870c12b1f265b2, Node: 0x1645..b7ea, addr: 58.96.74.212:9001, FINDNODE Request: distance: [245, 246, 244]
Oct 03 22:36:55.072 DEBG Session established with Node: 0x1645..b7ea, direction: Outgoing
Oct 03 22:36:55.245 TRCE Received a WHOAREYOU packet response. Source: Node: 0x1645..b7ea, addr: 58.96.74.212:9001
Oct 03 22:36:55.249 TRCE Sending Authentication response to node: Node: 0x1645..b7ea, addr: 58.96.74.212:9001 (External(RequestId([59, 135, 12, 18, 177, 242, 101, 178])))
Oct 03 22:36:55.250 TRCE Replaying active requests. Node: 0x1645..b7ea, addr: 58.96.74.212:9001, Some([182, 21, 66, 69, 225, 1, 225, 185, 238, 65, 162, 83])
Oct 03 22:36:55.250 TRCE Active request to be replayed. 0c9ea7b221a03aff, Node: 0x1645..b7ea, addr: 58.96.74.212:9001, PING: enr_seq: 448
Oct 03 22:36:55.250 DEBG Session established with Node: 0x1645..b7ea, direction: Outgoing
Oct 03 22:36:55.419 TRCE Received a WHOAREYOU packet response. Source: Node: 0x1645..b7ea, addr: 58.96.74.212:9001
Oct 03 22:36:55.420 TRCE Sending Authentication response to node: Node: 0x1645..b7ea, addr: 58.96.74.212:9001 (External(RequestId([12, 158, 167, 178, 33, 160, 58, 255])))
Oct 03 22:36:55.420 TRCE Replaying active requests. Node: 0x1645..b7ea, addr: 58.96.74.212:9001, Some([23, 228, 3, 99, 244, 82, 133, 251, 128, 201, 238, 134])
Oct 03 22:36:55.420 TRCE Active request to be replayed. 3b870c12b1f265b2, Node: 0x1645..b7ea, addr: 58.96.74.212:9001, FINDNODE Request: distance: [245, 246, 244]
Oct 03 22:36:55.420 DEBG Session established with Node: 0x1645..b7ea, direction: Outgoing
Oct 03 22:36:55.593 TRCE Received a WHOAREYOU packet response. Source: Node: 0x1645..b7ea, addr: 58.96.74.212:9001
Oct 03 22:36:55.597 TRCE Sending Authentication response to node: Node: 0x1645..b7ea, addr: 58.96.74.212:9001 (External(RequestId([59, 135, 12, 18, 177, 242, 101, 178])))
Oct 03 22:36:55.597 TRCE Replaying active requests. Node: 0x1645..b7ea, addr: 58.96.74.212:9001, Some([48, 31, 22, 136, 140, 235, 48, 105, 174, 255, 116, 247])
Oct 03 22:36:55.597 TRCE Active request to be replayed. 0c9ea7b221a03aff, Node: 0x1645..b7ea, addr: 58.96.74.212:9001, PING: enr_seq: 448
Oct 03 22:36:55.597 DEBG Session established with Node: 0x1645..b7ea, direction: Outgoing
Oct 03 22:36:55.768 TRCE Received a WHOAREYOU packet response. Source: Node: 0x1645..b7ea, addr: 58.96.74.212:9001
Oct 03 22:36:55.768 TRCE Sending Authentication response to node: Node: 0x1645..b7ea, addr: 58.96.74.212:9001 (External(RequestId([12, 158, 167, 178, 33, 160, 58, 255])))
Oct 03 22:36:55.768 TRCE Replaying active requests. Node: 0x1645..b7ea, addr: 58.96.74.212:9001, Some([241, 90, 55, 222, 212, 62, 126, 35, 91, 230, 95, 176])
Oct 03 22:36:55.768 TRCE Active request to be replayed. 3b870c12b1f265b2, Node: 0x1645..b7ea, addr: 58.96.74.212:9001, FINDNODE Request: distance: [245, 246, 244]
Oct 03 22:36:55.768 DEBG Session established with Node: 0x1645..b7ea, direction: Outgoing
Oct 03 22:36:55.942 TRCE Received a WHOAREYOU packet response. Source: Node: 0x1645..b7ea, addr: 58.96.74.212:9001
Oct 03 22:36:55.943 TRCE Sending Authentication response to node: Node: 0x1645..b7ea, addr: 58.96.74.212:9001 (External(RequestId([59, 135, 12, 18, 177, 242, 101, 178])))
Oct 03 22:36:55.943 TRCE Replaying active requests. Node: 0x1645..b7ea, addr: 58.96.74.212:9001, Some([112, 243, 120, 202, 245, 15, 56, 240, 197, 46, 174, 19])
Oct 03 22:36:55.943 TRCE Active request to be replayed. 0c9ea7b221a03aff, Node: 0x1645..b7ea, addr: 58.96.74.212:9001, PING: enr_seq: 448
Oct 03 22:36:55.943 DEBG Session established with Node: 0x1645..b7ea, direction: Outgoing
```


</p>
</details>

**Comment by AgeManning** [comment]
If that is used in the wild, it looks like the Node doesn't like our Auth response. 

If you use the master branch and see this node again, maybe it still behaves the same way. i.e it could be that node's fault.

**Comment by ackintosh** [comment]
Here is what I found after further investigation. These diagrams below assumes:

* Node1 attempts to send multiple requests(`id:1` and `id:2`) in parallel.
* Node2 doesn't send responses but only WHOAREYOU.

The sequence diagrams below show differences between `master` and `experimental` in terms of how they behave in that case.

## `master` branch

With `master` branch, the two requests result in failure (timeout). This is expected since Node2 doesn't respond requests.

```mermaid
sequenceDiagram
    participant Node1
    participant Node2

    Note left of Node1: Node1 attempts to send two requests in parallel
    Node1 ->> Node2: Random packet(id:1)
    Note over Node1: Insert Request(id:2) into `pending_requests`
    Node1 ->> Node1: Request(id:2)

    Node2 -->> Node1: WHOAREYOU(id:1)
    
    Note over Node1: New session
    Node1 ->> Node2: Handshake message(id:1)
    Note over Node2: Ignore Handshake message(id:1)

    Note over Node1: Request(id:1) timed out
    Note over Node1: Then Request(id:2) result in failure <br>due to the Request(id:1) timeout

    rect rgb(10, 100, 10)
    Note over Node1: Request(id:1) and Request(id:2) have been completed<br> in failure (timeout).
    end
```


## `experimental` branch

With `experimental` branch, the requests never complete. As illustrated below, the creation of sessions and the replay of requests continue in a loop.

```mermaid
sequenceDiagram
    participant Node1
    participant Node2

    Note left of Node1: Node1 attempts to send two requests in parallel
    Node1 ->> Node2: Random packet(id:1)
    Note over Node1: Insert Request(id:2) into `pending_requests`
    Node1 ->> Node1: Request(id:2)

    Node2 -->> Node1: WHOAREYOU(id:1)

    rect rgb(10, 10, 10)
    Note over Node1: New session
    Node1 ->> Node2: Handshake message(id:1)
    Note over Node2: Ignore Handshake message(id:1)
    Note over Node1: Send pending requests
    Node1 ->> Node2: Request(id:2)
    Node2 -->> Node1: WHOAREYOU(id:2)
    end

    rect rgb(10, 10, 10)
    Note over Node1: New session
    Node1 ->> Node2: Handshake message(id:2)
    Note over Node2: Ignore Handshake message(id:2)
    Note over Node1: Replay active requests
    Node1 ->> Node2: Request(id:1)
    Node2 -->> Node1: WHOAREYOU(id:1)
    end 

    rect rgb(10, 10, 10)
    Note over Node1: New session
    Node1 ->> Node2: Handshake message(id:1)
    Note over Node2: Ignore Handshake message(id:1)
    Note over Node1: Replay active requests
    Node1 ->> Node2: Request(id:2)
    Node2 -->> Node1: WHOAREYOU(id:2)
    end

    rect rgb(100, 10, 10)
    Note over Node1,Node2: The creation of sessions and the replay of requests <br>will continue in a loop, <br>alternating between Request(id:1) and Request(id:2).
    end
```

I think the cause of the problem is [here](https://github.com/sigp/discv5/blob/5e05a7d442c1ce621e53e5eba3d4842c70ed3c4d/src/handler/mod.rs#L957-L1007). On `Handler::replay_active_requests()` [we deletes the active requests to be replayed from `active_requests`](https://github.com/sigp/discv5/blob/5e05a7d442c1ce621e53e5eba3d4842c70ed3c4d/src/handler/mod.rs#L968-L975), and then on `Handler::send_request()` [the active requests are inserted into `active_requests` again](https://github.com/sigp/discv5/blob/5e05a7d442c1ce621e53e5eba3d4842c70ed3c4d/src/handler/mod.rs#L529). By doing this the request timeout of the active requests are reset on each `Handler::replay_active_requests()` calls. This is why the requests in the diagram never complete.

```mermaid
sequenceDiagram
    participant Node1
    participant Node2

    rect rgb(10, 10, 10)
    Note over Node1: New session
    Node1 ->> Node2: Handshake message(id:2)
    Note over Node2: Ignore Handshake message(id:2)
    rect rgb(100, 100, 10)
    Note over Node1: Replay active requests<br><br>Delete/Insert the active requests to be replayed.<br>*** So the request timeout(id:1) is reset. ***
    end
    Node1 ->> Node2: Request(id:1)
    Node2 -->> Node1: WHOAREYOU(id:1)
    end 
```

ackintosh:
@AgeManning @njgheorghita 
I have filed https://github.com/njgheorghita/discv5/pull/6 to fix [the problem](https://github.com/sigp/discv5/pull/200#issuecomment-1752383723) happening in the `experimental` branch. With this PR, the infinite loop of session creation and replaying requests is fixed. Requests of `Node1` will complete in failure as expected, as shown in the sequence diagram below. 

Also I have tested the fix with lighthouse and confirmed that works fine.

FYI: I have implemented a test plan ([sandbox](https://github.com/ackintosh/discv5-testground/pull/126)) of testground for this problem. Now we can easily reproduce the situation by mocking the behaviour of `Node2`.

```mermaid
sequenceDiagram
    participant Node1
    participant Node2

    Note left of Node1: Node1 attempts to send two requests(id:1 and id2)<br> in parallel
    Note right of Node2: Node2 doesn't send responses <br>but only WHOAREYOU

    Node1 ->> Node2: Random packet(id:1)
    Note over Node1: Insert Request(id:2) into `pending_requests`
    Node1 ->> Node1: Request(id:2)

    Node2 -->> Node1: WHOAREYOU(id:1)

    rect rgb(10, 10, 10)
    Note over Node1: New session
    Node1 ->> Node2: Handshake message(id:1)
    Note over Node2: Ignore Handshake message(id:1)
    Note over Node1: Send pending requests
    Node1 ->> Node2: Request(id:2)
    Node2 -->> Node1: WHOAREYOU(id:2)
    end

    rect rgb(10, 10, 10)
    Note over Node1: New session
    Node1 ->> Node2: Handshake message(id:2)
    Note over Node2: Ignore Handshake message(id:2)
    Note over Node1: Replay active requests
    Node1 ->> Node2: Request(id:1)
    Node2 -->> Node1: WHOAREYOU(id:1)
    end 

    rect rgb(10, 10, 10)
    rect rgb(10, 100, 10)
    Note over Node1: Handshake message(id:1) already sent
    Note over Node1: Request(id:1) failed
    Note over Node1: Request(id:2) failed
    end
    end
```

ackintosh:
After the fix has merged, I had run this on a live network and verified it works as expected.

AgeManning:
Ok, @ackintosh we think this ready for a merge?

I'll do one last quick review if so.

ackintosh:
@AgeManning Yes, please go ahead with the final review. 🙂

AgeManning:
This seems like it's been well tested. I think the best approach now is to merge it in. We're hoping to get better visibility on errors that occur is discovery, with  https://github.com/sigp/discv5/pull/218

I think we should make a release with both of these PRs in and we will be able to test at scale in Lighthouse also.

AgeManning (approved):

AgeManning:
@divagant-martian - Want to do a final review, happy with a merge then live test?

AgeManning:
As soon as the review is approved lets merge this :)

divagant-martian (approved):
I'm mostly happy with the current state. Looks like the happy paths are appropriately handled. There are some places where I'd like to do some changes: Active requests can probably be changed to use a `Smallvec` instead of a `Vec` with a reasonable low value. Some places need a bit of love in handling the unhappy paths since we do things like this:
```
fn operation_that_should_be_atomic() {
    op1_that_works(); // this change remains even if op2 fails
    op2_that_fails()?;
}
```

but in all fairness the affected code already required some love.

That being said, I don't want to delay this further.  The base is good, now time to battle test it.

@AgeManning *please note that there is one required changed before merging*, since there is an unwrap in the code

src/handler/mod.rs
```diff
@@ -893,51 +914,100 @@ impl Handler {
...
+                    if let Some(nonce) = message_nonce.as_ref() {
+                        req.packet().message_nonce() != nonce
+                    } else {
+                        true
+                    }
+                })
+            {
+                let new_packet = session
+                    .encrypt_message::<P>(self.node_id, &request_call.encode())
+                    .unwrap();
```
divagant-martian, line 989: I kinda believe this but we don't really use `unwrap` in prod code (release builds). Can you please change this to handle the error or at least change it to an `expect` with the right explanation?

This is my only requirement before merging

AgeManning: @njgheorghita - I had a quick look through the AES library we use. From what I can tell this function can fail if the size of the messages are over a specific limit. I was looking at version 0.9. I haven't looked at future versions. 

I don't think it's reasonable to assume that in the current version and all possible future versions, that every possible message we try to encrypt will be successful. If there is an event that it's not successful, I dont think we should crash the entire application. 

Failing to encrypt a message is not a catastrophic failure imo. I think it's fine to just log a CRIT message and continue running. 

In which case we shouldn't use an `unwrap()` or `expect()` rather check the failure and if it fails log a crit. 

What do you think?

njgheorghita: Some thoughts...

- We're only dealing with requests retrieved from the `active_requests` store. IIUC this implies a pretty strong guarantee that the encryption will be successful, since the request has already been successfully encrypted when it originated. If this is accurate, then `expect` might actually expose an error in the request-handling logic - though, I'm unsure whether or not that is worth crashing...
- If the encryption fails, do we want to do more than just log a CRIT message? Maybe it's worth removing the request entirely from the `active_requests` store. Although... maybe it's alright to just leave the (unable to encrypt) active request as is and let it simply timeout? 

> Failing to encrypt a message is not a catastrophic failure imo. 

Ultimately, I agree with this. So I'm leaning towards logging a message. The question I would need to think more about (it's been a while since I've been intimate with this code 😅 ) is whether or not we need to handle un-encryptable requests? Or if it's ok to just log the message and let the request timeout

divagant-martian: The reason why I more or less believed this could be true is because we are replaying. In that scenario, the appropriate message in `expect` is *why* this is infallible, not simply the error occurred. Maybe could have been more clear about it.

But it seems to be replaying might no be enough guarantee from @AgeManning 's comment? In that case the error needs to be handled. Here that potentially means bobbling up the error to the application. You should look into how failed requests and failed encryption in particular is handled in other parts of the code to get a good idea of what to do in this scenario.

Also, yes to logging the error (there is no CRIT, just ERROR)

> Or if it's ok to just log the message and let the request timeout

Would need to check but I assume we try to guarantee that if a request times out then it is active, removing from one place and not the other would most likely break that invariant.

ackintosh:
https://github.com/sigp/discv5/releases/tag/v0.4.0 🎉

PR #204, opened 2023-07-22
fix entry docs

divagant-martian:
## Description

- Docs were using explicit references instead of relying on cargo finding them. This leads to broken links that `rustdoc` does not catch (because it trusts us), main one was the `Session`, which is private.
- Reword some descriptions to point to concrete modules and structs.
- Remove any mention to advertisement since it was removed.
- Remove some attributes that have not been necessary for a long time

## Notes & open questions

From reading the docs, tempted to remove the `Discv5` prefix as it was done in `rust-libp2p`. Would look cleaner internally and it seems unnecessary externally

## Change checklist

- [x] Self-review
- [ ] Tests if relevant

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,57 +1,52 @@
-#![warn(rust_2018_idioms)]
 #![deny(rustdoc::broken_intra_doc_links)]
-#![cfg_attr(docsrs, feature(doc_cfg))]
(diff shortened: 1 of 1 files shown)


AgeManning (approved):

AgeManning:
Yeah. Removing the prefix is probably a good idea.

PR #205, opened 2023-07-22
Add serde to Pong struct

KolbyML:
## Description

Add serde to Pong struct

## Notes & open questions

No rush to get this out, I just think this could be useful

## Change checklist

- [x] Self-review
- [ ] Tests if relevant

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ hex = "0.4.3"
 fnv = "1.0.7"
 arrayvec = "0.7.2"
 rand = { version = "0.8.4", package = "rand" }
+serde = { version = "1.0.110", features = ["derive"], optional = true }
 socket2 = "0.4.4"
 smallvec = "1.7.0"
 parking_lot = "0.11.2"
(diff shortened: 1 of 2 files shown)


divagant-martian:
is there a use case for this? Otherwise I don't think it's of use to include. Not against the change on itself -it seems simple and well implemented-, but there are plenty types that could derive `Serialize` and `Deserialiaze` that don't do since there is no need. So, why pong? it's particularly ephemeral

KolbyML:
I want to use the struct as the return type for one of our jsonrpsee endpoints which is calls send_ping and returns Pong. I can always just copy paste this struct then .into() what I get from sigp/discv5, but I thought it was clearer and less duplicate code to just add serde here.

divagant-martian:
could you please check what other public types trin would use in the json rpc to do this in a batch? Easier to understand as a future reader of the git history if they are all added together instead of one by one

AgeManning:
Yeah I agree. If we are going to implement a serialize/deserialze trait on our RPC, we should not do it ad-hoc. If we are going to implement on one, we should at least implement it on all.

divagant-martian:
closing this as stale

Issue #207, opened 2023-08-02
Encoded RPC Message bytes are not same


armaganyildirak:
When RPC Message was [decoded](https://github.com/sigp/discv5/blob/a9f1e99321aec746fb9d6e8df889aa515a5e1254/src/rpc.rs#L304) and [encoded](https://github.com/sigp/discv5/blob/a9f1e99321aec746fb9d6e8df889aa515a5e1254/src/rpc.rs#L297) these bytes ```[6, 193, 0, 75, 252]```,  Encoding message returns different result ```[6, 194, 0, 75]```

When the message is decoded, The message is ```Response(Response { id: RequestId([0]), body: Talk { response: [75] } })```

divagant-martian:
I did a bit of digging,

first I checked the spec against the code and at least to my eyes, encoding and decoding looks legit. Maybe it could be polished but it seems functionally correct to me.

Then I also did this test
```rust
        let data = [6, 193, 0, 75, 252];
        let msg = Message::decode(&data).unwrap();

        let data2 = [6, 194, 0, 75];
        let msg2 = Message::decode(&data2).unwrap();

        assert_eq!(msg, msg2);
```

It seems to be both representations produce the same message, so I'm no longer sure if this is an issue. It depends on rlp producing unique representations. And I'm not sure that's true

weipin:
[ Looked at this VERY briefly -- there are things I could be missing ]

There could be bugs -- either the RLP crate is misused, or the RLP crate 
itself fails to report certain errors.

Two issues: A) mismatch of "the number of list items" isn't reported. B) trailing
"garbage" of list items isn't reported.

Extending @divagant-martian 's code a little bit (added `data3`):

```
    let data = [6, 193, 0, 75, 252];
    let msg = Message::decode(&data).unwrap();

    let data2 = [6, 194, 0, 75];
    let msg2 = Message::decode(&data2).unwrap();

    let data3 = [6, 194, 0, 75, 252];
    let msg3 = Message::decode(&data3).unwrap();

    assert_eq!(msg, msg2);
    assert_eq!(msg, msg3);
```

`data` demonstrates issue A, `data3` demonstrates issue B. Both should fail and
return proper errors.

Here is a breakdown of decoding `data2`, the seemingly correct one.

Decoding `[6, 194, 0, 75]`:
  * `6`: message type for TALKRESP Response (0x06)
[...]

PR #208, opened 2023-08-03
fix rpc decoding to reject extra data

divagant-martian:
## Description

addresses #207 
verify that received data does not contain extra bytes when doing rlp encoding.

## Notes & open questions

This is (I think) appropriate workaround to what I would consider either a bug or an api shortcoming in `rlp`

## Change checklist

- [x] Self-review
- [x] Tests if relevant

### DIFF
--- a/src/rpc.rs
+++ b/src/rpc.rs
@@ -12,7 +12,6 @@ impl From<RequestId> for Vec<u8> {
         id.0
     }
 }
-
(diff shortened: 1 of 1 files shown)


AgeManning (approved):
This looks alright to me. 

We probably need to run some tests to see if there is any bad rlp out in the wild

PR #216, opened 2023-09-29
Use zero ports in tests

### LINKED ISSUE #215: Use zero ports in tests
Similarly to sigp/lighthouse#4704, some of our tests race on port number allocation. The main offenders are:

 - https://github.com/sigp/discv5/blob/299638d0336896625372e496c10e2840b3cfbfb8/src/service/test.rs#L102
 - https://github.com/sigp/discv5/blob/299638d0336896625372e496c10e2840b3cfbfb8/src/service/test.rs#L169

These both hardcode UDP ports (like 10001 and 10002). Locally, all it takes me is 3-4 runs of the test suite to induce failures due to `AddrInUse`. The specific failure case is always a panic when the port can't be allocated:

```
thread 'service::test::test_connection_direction_on_inject_session_established' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 98, kind: AddrInUse, message: "Address already in use" }', src/service/test.rs:58:14
```

jmcph4:
## Description

Closes #215.

## Notes & open questions

N/A

## Change checklist

- [x] Self-review
- [x] ~~Documentation updates if relevant~~
- [x] ~~Tests if relevant~~

### DIFF
--- a/src/service/test.rs
+++ b/src/service/test.rs
@@ -19,6 +19,9 @@ use parking_lot::RwLock;
 use std::{collections::HashMap, sync::Arc, time::Duration};
 use tokio::sync::{mpsc, oneshot};
 
+/// Default UDP port number to use for tests requiring UDP exposure
+pub const DEFAULT_UDP_PORT: u16 = 0;
+
 fn _connected_state() -> NodeStatus {
     NodeStatus {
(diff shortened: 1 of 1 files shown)


divagant-martian (approved):
LGTM!

PR #218, opened 2023-10-04
Expand metrics for errors and warnings

### LINKED ISSUE #213: Add Error Metric
It would be useful to add a warning and error metric to the global metrics. 

This way if an error or warning occurs we increment these counts. Often in Lighthouse, we dont emit logs for discovery and a variety of errors and warnings can go unnoticed. 

If we had these in the metrics we could just monitor metrics for discovery to observe if anything is going wrong. 
I think it would be good to add the following metrics:
- Error count
- Warning count
- Error Type - Give some string context as to which type of error this is. As the debug logs are not saved and we see a large error/warning count it would be good to provide some context into what these errors are.

jmcph4:
## Description

Closes #213.

## Notes & open questions

This PR expands the metrics collected by the Discv5 service to include errors and warnings that occur throughout the operation of the service. This is achieved by the introduction of a new type, `ErrorMetrics`, which stores four elements:

 - Total number of errors
 - Total number of warnings
 - Mapping of specific errors to their associated counts
 - Mapping of specific warnings to their associated counts

A few technical notes:

 - Cumulative counts are stored *in addition to* individual counts to avoid having to load each `AtomicUsize` in a manner such as `self.errors.read().values().map(|at| at.load(Ordering::Relaxed).sum()`. The additional memory overhead is considered negligible as is the added locking overhead
 - Errors and warnings are uniquely identified solely by strings. Presently, these are provided by named constants within the `metrics` module. Assigning unique numeric identifiers was considered but would add complexity with little benefit (these will eventually be turned back into *some* string representation when these hit Grafana/Prometheus, anyway).

Notes for reviewers:

 - I'm very open to the names of these error/warning strings changing to be more descriptive/clear, but once again, they are still largely arbitrary
 - With regards to testing, these changes are particularly challenging to test as they exclusively target failure modes of the discovery service. Manual testing, by executing [the various examples](https://github.com/sigp/discv5/tree/master/examples) in the repository, was performed in order to check for basic defects (such as panics, deadlocks, fatal errors, etc.). Additionally, I plumbed this branch into Lighthouse mainline and tested by polling the metrics endpoint (this won't yield any of the metrics in this PR as they need to be added to LH separately, but this nonetheless validates that the discovery service is functioning correctly even with the changes in this PR on a live BN).

## Change checklist

- [x] Self-review
- [x] Documentation updates if relevant (by way of doc comments)
- [x] ~~Tests if relevant~~

### DIFF
--- a/src/handler/active_requests.rs
+++ b/src/handler/active_requests.rs
@@ -1,3 +1,5 @@
+use crate::metrics::{NO_MATCHING_NONCE, NO_MATCHING_REQ_CALL};
+
 use super::*;
 use delay_map::HashMapDelay;
 use more_asserts::debug_unreachable;
@@ -39,6 +41,7 @@ impl ActiveRequests {
                 None => {
                     debug_unreachable!("A matching request call doesn't exist");
                     error!("A matching request call doesn't exist");
+                    let _ = &METRICS.error(NO_MATCHING_REQ_CALL);
                     None
                 }
             },
@@ -58,6 +61,7 @@ impl ActiveRequests {
                     None => {
                         debug_unreachable!("A matching nonce mapping doesn't exist");
                         error!("A matching nonce mapping doesn't exist");
+                        let _ = &METRICS.error(NO_MATCHING_NONCE);
                         None
                     }
                 }
--- a/src/handler/mod.rs
+++ b/src/handler/mod.rs
@@ -30,6 +30,7 @@ use crate::{
     config::Config,
     discv5::PERMIT_BAN_LIST,
     error::{Error, RequestError},
+    metrics::SESS_GENERATE_FAIL,
     packet::{ChallengeData, IdNonce, MessageNonce, Packet, PacketKind, ProtocolIdentity},
     rpc::{Message, Request, RequestBody, RequestId, Response, ResponseBody},
     socket,
@@ -668,6 +669,7 @@ impl Handler {
             Ok(v) => v,
             Err(e) => {
                 error!("Could not generate a session. Error: {:?}", e);
+                let _ = &METRICS.error(SESS_GENERATE_FAIL);
                 self.fail_request(request_call, RequestError::InvalidRemotePacket, true)
                     .await;
                 return;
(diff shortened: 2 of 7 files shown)


jmcph4:
Discussed this out-of-band with @AgeManning and we agreed to pivot the design of how Discv5 metrics are to work. As such, this PR is going through a refactor.

AgeManning:
I believe the state should be like this:

We keep this PR. We add metrics to discv5 (and to libp2p) that gives us a count of errors and their severity. I.e number of crits, errors, warns. 
In conjunction we implement: https://github.com/sigp/lighthouse/issues/4954 which allows us to store seperate log files for discv5 and libp2p. 

The idea then would be that we can check our metrics dash to know if there are any underlying problems and we can then go to the log files to investigate further. 


This PR needs to simply count the error logs.

AgeManning:
A simple way of doing this, it to replace the warn!, error! and crit! macros with ones that increment a metric count. No further could would need to change. 

Perhaps @armaganyildirak may be interested in this.

divagant-martian:
I expressed this via other channels but will leave this here for the record. 

I don't think this is the right approach to solve this issue. If we end up doing this it should be as a last resort since doing this is highly polluting to the code and very hard to maintain. I proposed a different approach: to build a `tracing::Subscriber` that records metrics in LH itself.

Since I have not seen this implemented yet, I'll give it a go and report back.

In case I find the proposed approach not doable or equally hard to maintain, we could continue with providing these metrics from discv5. However, this should be done with macros, per log level, and ignoring the message of the log to reduce the polluting and maintenance work. (@AgeManning 's [last comment](https://github.com/sigp/discv5/pull/218#issuecomment-1831447937), basically) Both scenarios diverge enough from the current PR to still consider it reasonable to build on top of this one even in the second scenario.

This being said, I'm closing this PR. Second scenario should start a new one if that's where we land

divagant-martian:
alternative here: https://github.com/sigp/lighthouse/pull/4979 it ended up being fairly simple

PR #219, opened 2023-10-19
update dependencies

jxs:
## Description

and unify libp2p dep

## Change checklist

- [x] Self-review
- [ ] Tests if relevant

### DIFF
--- a/src/socket/filter/mod.rs
+++ b/src/socket/filter/mod.rs
@@ -6,8 +6,8 @@ use enr::NodeId;
 use lru::LruCache;
 use std::{
     collections::HashSet,
-    convert::TryInto,
     net::{IpAddr, SocketAddr},
+    num::NonZeroUsize,
     sync::atomic::Ordering,
     time::{Duration, Instant},
 };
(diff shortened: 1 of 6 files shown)


src/socket/filter/mod.rs
```diff
@@ -67,8 +68,8 @@ impl Filter {
                 expected_packets_per_second,
                 METRICS.moving_window,
             ),
-            known_addrs: LruCache::new(KNOWN_ADDRS_SIZE),
-            banned_nodes: LruCache::new(BANNED_NODES_SIZE),
+            known_addrs: LruCache::new(KNOWN_ADDRS_SIZE.try_into().unwrap()),
+            banned_nodes: LruCache::new(BANNED_NODES_SIZE.try_into().unwrap()),
```
divagant-martian, line 72: Neither of those should be a runtime error (if it was then we should be using `expect` instead of `unwrap`)
you can instead change the declarations to 
```rust
const KNOWN_ADDRS_SIZE: NonZeroUsize = match NonZeroUsize::new(500) {
    Some(non_zero) => non_zero,
    None => panic!("clearly non zero"),
};
```
to get a compile time check via const-panic

divagant-martian (changes requested):
Thanks for the updates! Changes are not compiling rn, could you check it out @jxs?

src/socket/filter/mod.rs
jxs: thanks Diva, yeah I assumed not being able to `unwrap` meant one couldn't `panic!()` in `const fn`s but it's possible, I used `unreachable!()` intead of `panic!()`inc ptal.

divagant-martian (approved):
perfect, thanks

PR #234, opened 2024-02-13
Merge `master` into `discv5.2`

ackintosh:
## Description

Merge lastest `master` branch into `discv5.2`.

## Notes & open questions

I have updated the `Config::session_cache_capacity` type from usize to NonZeroUsize in 6d4fce2f05f51823021f971be41d4de379c313ad, for the following reasons:

- `Config::session_cache_capacity` is utilized in [Handler's session](https://github.com/ackintosh/discv5/blob/6d4fce2f05f51823021f971be41d4de379c313ad/src/handler/mod.rs#L332) (LruTimeCache) and in [`Nat::new_peer_latest_relay_cache`](https://github.com/ackintosh/discv5/blob/238dbc0d6d390f249157ce32027431577059efb9/src/handler/nat.rs#L59) (LruCache).
- The type of `Nat::new_peer_latest_relay_cache` is LruCache and the argument for `LruCache::new` is NonZeroUsize

## Change checklist

- [x] Self-review
- [x] Tests if relevant

### DIFF
--- a/src/service.rs
+++ b/src/service.rs
@@ -29,7 +29,7 @@ use crate::{
     query_pool::{
         FindNodeQueryConfig, PredicateQueryConfig, QueryId, QueryPool, QueryPoolState, TargetKey,
     },
-    rpc, Discv5Config, Discv5Event, Enr, IpMode,
+    rpc, Config, Enr, Event, IpMode,
 };
 use delay_map::HashSetDelay;
 use enr::{CombinedKey, NodeId};
@@ -40,6 +40,7 @@ use parking_lot::RwLock;
 use rpc::*;
 use std::{
     collections::HashMap,
+    convert::TryInto,
     net::{IpAddr, SocketAddr},
     sync::Arc,
     task::Poll,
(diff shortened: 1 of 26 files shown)


src/service.rs
```diff
@@ -1507,13 +1508,14 @@ impl Service {
...
                         if !nodes_response.received_nodes.is_empty() {
-                            warn!(
-                                "NODES Response failed, but was partially processed from: {}",
-                                active_request.contact
-                            );
+                            let node_id = active_request.contact.node_id();
+                            let addr = active_request.contact.socket_addr();
+                            let received = nodes_response.received_nodes.len();
+                            let expected = distances.len();
+                            warn!(%node_id, %addr, %error, %received, %expected, "FINDNODE request failed with partial results");
```
emhane, line 1518: isn't this wording a bit misleading? distances and nodes don't map 1-1. we can receive more than one node in the response at any distance in the requested range of distances.

```suggestion
                            let received = nodes_response.received_nodes.len();
                            warn!(%node_id, %addr, %error, %received, ?distances, "FINDNODE request failed with partial results");
```

Issue #239, opened 2024-03-12
errors after update, related to alloy_rlp


pedrohba1:
Not sure if this is known already, but here is the output (v0.4.1)

```rs
error[E0432]: unresolved import `enr::EnrError`
  --> /home/bufulin/.cargo/registry/src/index.crates.io-6f17d22bba15001f/discv5-0.4.1/src/discv5.rs:26:24
   |
26 | use enr::{CombinedKey, EnrError, EnrKey, NodeId};
   |                        ^^^^^^^^ no `EnrError` in the root
error[E0277]: the trait bound `T: alloy_rlp::encode::Encodable` is not satisfied
   --> /home/bufulin/.cargo/registry/src/index.crates.io-6f17d22bba15001f/discv5-0.4.1/src/discv5.rs:443:26
    |
443 |             .insert(key, value, &self.enr_key.read())
    |              ------      ^^^^^ the trait `alloy_rlp::encode::Encodable` is not implemented for `T`
    |              |
    |              required by a bound introduced by this call
    |
note: required by a bound in `enr::Enr::<K>::insert`
   --> /home/bufulin/.cargo/registry/src/index.crates.io-6f17d22bba15001f/enr-0.10.1/src/lib.rs:486:22
    |
486 |     pub fn insert<T: Encodable>(
    |                      ^^^^^^^^^ required by this bound in `Enr::<K>::insert`
help: consider further restricting this bound
    |
436 |     pub fn enr_insert<T: rlp::Encodable + alloy_rlp::encode::Encodable>(
    |                                         ++++++++++++++++++++++++++++++

   Compiling jsonrpsee v0.20.3
error[E0277]: the trait bound `enr::Enr<CombinedKey>: Encodable` is not satisfied
   --> /home/bufulin/.cargo/registry/src/index.crates.io-6f17d22bba15001f/discv5-0.4.1/src/packet/mod.rs:176:35
    |
176 |                 let node_record = enr_record.as_ref().map(rlp::encode);
    |                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Encodable` is not implemented for `enr::Enr<CombinedKey>`
    |
    = help: the following other types implement trait `Encodable`:
              bool
              usize
              u8
              u16
              u32
              u64
              u128
              Box<T>
            and 7 others
note: required by a bound in `rlp::encode`
[...]

PR #240, opened 2024-03-15
Set precise enr version

emhane:
## Description

Sets precise enr dep v0.10.0. Closes https://github.com/sigp/discv5/issues/239.

## Change checklist

- [x] Self-review
- [ ] Tests if relevant


AgeManning:
Thanks for this interim solution. 

In general we still want to allow ENR patch updates without having to make releases of discv5. 

This error came from my mistake in semver where the changes in the 10.1 release were breaking but I didn't notice. It's been fixed in the ENR crate. 

In principle, it shouldn't happen in the future ;)

Issue #243, opened 2024-03-26
`entered unreachable code: Stored ENR is not contactable`


emhane:
Goes through unreachable path.

https://github.com/sigp/discv5/blob/6707bcf3da350ff273208bb38dc29acc4cc4850c/src/service.rs#L598

Have a hunch this has to do with adding multiaddresses, since this is how I'm adding enodes in my pr, entry point here

https://github.com/paradigmxyz/reth/blob/993036014eb6c3cb51e308b885339e483771dbb4/crates/net/network/src/config.rs#L553

```
thread 'tokio-runtime-worker' panicked at /home/ubuntu/.cargo/git/checkouts/discv5-7af694118de5b208/04ac004/src/service.rs:598:29:                                                                                                                   
internal error: entered unreachable code: Stored ENR is not contactable. enr:-Le4QAAHJBTIJ1hP4eLJgvFFJ2E29tIVz9LEgQeACShg3sMxGgt-zi7bVlaHdEICENFuir0QAxMuW_Cj_39hhTjPUZSCB3-HYXR0bmV0c4hgAAAAAAAAAIRldGgykGmuDpkFAXAA__________-CaWSCdjSEcXVpY4IjM4lz
ZWNwMjU2azGhA7gc3n5LM3gKW91FEtG_4J_a-lgwJxot6tLQtHnmomEOiHN5bmNuZXRzDIN0Y3CCIzw                                                                                                                                                                      
stack backtrace:                                                                                                                                                                                                                                     
   0: rust_begin_unwind                                                                                                                                                                                                                              
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/panicking.rs:645:5                                                                                                                                                   
   1: core::panicking::panic_fmt                                                                                                                                                                                                                     
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/panicking.rs:72:14                                                                                                                                                  
   2: discv5::service::Service::handle_rpc_request                                                                                                                                                                                                   
             at /home/ubuntu/.cargo/git/checkouts/discv5-7af694118de5b208/04ac004/src/service.rs:598:29                                                                                                                                              
   3: discv5::service::Service::start::{{closure}}                                                                                                                                                                                                   
             at /home/ubuntu/.cargo/git/checkouts/discv5-7af694118de5b208/04ac004/src/service.rs:385:33                                                                                                                                              
[...]

PR #246, opened 2024-04-02
Prevent storing non-contactable ENRs

AgeManning:
ENR updates could allow for storing non-contactable ENRs. 

This PR handles some edge cases where we could update old ENRs with newer ones which are no longer contactable.

### DIFF
--- a/src/service.rs
+++ b/src/service.rs
@@ -886,7 +886,10 @@ impl Service {
                                 };
                                 self.send_rpc_request(active_request);
                             }
-                            self.connection_updated(node_id, ConnectionStatus::PongReceived(enr));
+                            // Only update the routing table if the new ENR is contactable
+                            if self.ip_mode.get_contactable_addr(&enr).is_some() {
+                                self.connection_updated(node_id, ConnectionStatus::PongReceived(enr));
+                            }
                         }
                     }
                 }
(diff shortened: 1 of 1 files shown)


AgeManning:
I think this should resolve #243

src/service.rs
```diff
@@ -1221,7 +1230,22 @@ impl Service {
...
+                    kbucket::Entry::Present(entry, _) if entry.value().seq() < enr.seq() => {
+                        entry.remove()
+                    }
+                    kbucket::Entry::Pending(mut entry, _) => {
+                        if entry.value().seq() < enr.seq() {
+                            entry.remove()
+                        }
+                    }
+                    _ => {}
+                }
```
jxs, line 1245: nit:
is there be a situation for legit storing a non contactable ENR?
If not can we (instead) log an error and have a `debug_unreachable` if it exists in `kbuckets` like in
https://github.com/sigp/discv5/blob/6707bcf3da350ff273208bb38dc29acc4cc4850c/src/service.rs#L598

jxs (approved):
LGTM age

ackintosh (approved):
LGTM!

src/service.rs
AgeManning: The entry here is about finding a new ENR. 

So we could have an ENR that is contactable and we store it in the routing table. Then later, we find they have changed their ENR to something non-contactable and we see an updated version. Because the key in the table is the node-id, it can be that the earlier version was contactable but the new one is not. So we remove the old one here and don't store the new one. 

I can't wrap this in a debug only mode, because this is a valid case. 

For the sake of speed, I"ll merge this PR and if my analysis here is wrong we can make a future PR to add this in.

jxs: Thanks for clarifying age, makes sense

PR #250, opened 2024-04-16
Optionally notify app of unverifiable enr

emhane:
## Description

Discv5 is a great library, but to use it with EL some quirks have been needed. The major one being, notifying the app of discovered peers that advertise a socket that doesn't verify against their actual socket.

How a node behaves before it has discovered its WAN reachable socket, is not spec'd out. The go implementation advertises localhost in the ENR, if no WAN reachable socket is set, since it's the least usable address the node can advertise (as opposed to useless no address). This doesn't fit with the rust implementation, which verifies ENRs against their source to mitigate DoS attacks against the advertised address.

## Notes & open questions

Adding this event breaks the API. To allow for adding more events in the future without breaking the api, the `non_exhaustive` attribute is added to `Event`.

## Change checklist

- [x] Self-review
- [ ] Tests if relevant

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -98,6 +98,9 @@ pub struct Config {
 
     /// Configuration for the sockets to listen on.
     pub listen_config: ListenConfig,
+
+    /// Flag that enables notifying the application layer of unverifiable ENRs, peers that are denied sessions. Defaults to `false`.
+    pub notify_unverifiable_enr: bool,
 }
 
 #[derive(Debug)]
@@ -143,6 +146,7 @@ impl ConfigBuilder {
             ban_duration: Some(Duration::from_secs(3600)), // 1 hour
             executor: None,
             listen_config,
+            notify_unverifiable_enr: false,
         };
 
         ConfigBuilder { config }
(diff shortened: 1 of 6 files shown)


src/config.rs
```diff
@@ -98,6 +98,9 @@ pub struct Config {
 
     /// Configuration for the sockets to listen on.
     pub listen_config: ListenConfig,
+
+    /// Flag that enables notifying the application layer of unverifiable ENRs, peers that are denied sessions. Defaults to `false`.
+    pub notify_unverifiable_enr: bool,
```
AgeManning, line 103: Appreciate the concern to make it optional to add EL specific features, 

But I think in this case, its just an event. The event-stream is optional for users, so they can just not listen to the event stream if they don't want. 

I dont think we need a config param, lets just make it always on, which saves a bit of code and complexity.

AgeManning (commented):
Nice! LGTM

AgeManning:
Once we get this down, we can make a new minor release, which will also target a new ENR minor release

AgeManning (approved):

PR #256, opened 2024-04-29
chore(deps): replace libp2p dep with libp2p-identity and multiaddr

### LINKED ISSUE #253: reduce libp2p dependencies
the libp2p feature pulls in libp2p entirely:

https://github.com/sigp/discv5/blob/44051ef77146003adb100cf87952da42a68e4174/Cargo.toml#L17-L17

but only identity and MultiAddr are used

https://github.com/sigp/discv5/blob/44051ef77146003adb100cf87952da42a68e4174/src/node_info.rs#L7-L11

which are available as standalone crates:

`multiaddr` and `libp2p-identity`

mattsse:
## Description

Closes #253

Reduces dependencies from 202 to 182 if libp2p feature is enabled by only importing required crates: `multiaddr` + `libp2p-identity` 

## Change checklist

- [x] Self-review
- [ ] Tests if relevant

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,8 @@ exclude = [".gitignore", ".github/*"]
 [dependencies]
 enr = { version = "0.12", features = ["k256", "ed25519"] } 
 tokio = { version = "1", features = ["net", "sync", "macros", "rt"] }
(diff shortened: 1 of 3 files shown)


divagant-martian:
#255 was added just about a minute earlier and addresses this already

PR #265, opened 2024-09-18
Respond to requests from peers with invalid ENRs

AgeManning:
It can be the case that peers have invalid ENRs (i.e report socket addresses that do not match the socket src of the packet). 

Previously discovery would drop packets from these peers and log a warning. This PR shifts this logic to instead respond to the src socket but exclude the ENR from reaching our routing table. 

We will not advertise these peers, but we will respond to them. 

I think this change will improve overall connectivity.

There are some legitimate cases where an ENR can be invalid, such as the external IP address changing or some funky NATs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,11 +80,6 @@ A simple example of creating this service is as follows:
 
 # Addresses in ENRs 
 
-This protocol will drop messages (i.e not respond to requests) from peers that
-advertise non-contactable address in their ENR (e.g `127.0.0.1` when connecting
-to non-local nodes). This section
-explains the rationale behind this design decision.
-
 An ENR is a signed record which is primarily used in this protocol for
 identifying and connecting to peers. ENRs have **OPTIONAL** `ip` and `port`
 fields.
@@ -127,8 +122,5 @@ This is done in the following way:
 3. If a peer connects to us with an ENR that specifies an IP address that does
    not match the src socket it connects to us on (e.g `127.0.0.1`, or
    potentially some internal subnet IP that is unreachable from our current
-   network) we consider this peer malicious/faulty
-   and drop all packets. This way we can efficiently drop peers that may try to
-   get us to send messages to arbitrary remote IPs, and we can be sure that all
-   ENRs in our routing table are contactable (at least by our local node at
-   some point in time).
+   network) we consider this ENR invalid and will not add it to our routing
+   table. However we will still respond to discovery requests.
--- a/src/handler/mod.rs
+++ b/src/handler/mod.rs
@@ -74,12 +74,6 @@ use session::Session;
 // seconds).
 const BANNED_NODES_CHECK: u64 = 300; // Check every 5 minutes.
 
-// The one-time session timeout.
-const ONE_TIME_SESSION_TIMEOUT: u64 = 30;
-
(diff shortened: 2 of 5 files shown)


ackintosh:
I'm testing this change with the [`ip-change`](https://github.com/ackintosh/discv5-testground?tab=readme-ov-file#ip-change) test case. I'll let you know when it's done.

AgeManning:
Awesome. Thanks @ackintosh :pray:

ackintosh:
@AgeManning During the testing, I noticed that the FINDNODE request (which is initiated from [here](https://github.com/sigp/discv5/blob/8d20f811553646d2279e38b247334aee92d3ff8e/src/service.rs#L598)) is timing out. Please see the attached image below. What do you think?

image

ackintosh:
For reference, the issue I mentioned above doesn’t happen in the current implementation, which uses a one-time session:

image

AgeManning:
@ackintosh 

Awesome!! Thanks for catching this! 

I was thinking in the transition its probably best to simply remove the transitioning peer from the local routing table. There is likely some time between us noticing it has changed its IP and it actually adjusting its IP. In the meantime we probably should prioritize better peers. 

I suspect also if we remove it from the table, when it does eventually update, it can then have a second chance of getting back in there. 

I've made this change. I'm hoping it now fixes the issue you've found. 

Thanks again!

ackintosh:
@AgeManning I have tested it again. The updated ENR doesn't get back into Node B's routing table unless the session has expired. 🤔 

image

AgeManning:
Hey @ackintosh 

Nice. Yeah, I'm trying to think of any better way to improve on this behaviour. There's a number of things we want and facts we have to deal with:
- We want to maintain the connection and respond to invalid ENRs now, so we dont want to terminate the session now in the process of a peer changing its IP
- While its ENR is invalid, we dont want to advertise it and therefore don't want it in the routing table
- There is some indeterminate time before the node updates its ENR (it might not even update it). 

Therefore, I dont think we should: 
-  Download every ENR we see that could be new on an established connection. Especially if we dont have them in our routing table, then we don't care about the ENR update
- Keep the ENR in the routing table while we know its non-contactable
- Drop and re-establish connections in hope that the ENR is now correct. 


Given these, I can't think of an efficient and useful way that we can re-add this ENR to the table without the session expiring. We could keep track of expired node-ids and re-download ENRs if their sequence increases, but we could be downloading lots of ENRs with no benefit. Also the code complexity would increase a bit for this edge case. 

I'm open to ideas of how to do this nicely, but currently I'm thinking to just leave this behaviour. In a real DHT, we have lots of peers. If we respond to everyone now and we kick out non-contactable ones, it should make the whole thing cleaner, i'd imagine.

ackintosh:
@AgeManning The idea that came to mind was expanding the use of one-time sessions not only for PING, but also for all other messages. If the ENR is invalid, then remove it from the routing table and respond to the peer using a one-time session. This gives the ENR a chance to get back to our routing table when the peer updates it, while still allowing us to respond to the peer.

However, this might increase the code complexity a bit, and yeah, this might not be a good solution, especially given that the peer might not update its ENR.

ackintosh (approved):
Looks good to me!

PR #266, opened 2024-10-08
More lenient ipv6 auto-update

AgeManning:
## Description

On IPv4 dominated networks, local routing tables can be fully populated with ipv4 nodes limiting the number of ipv6 ping/pong responses. This can prevent the enr-auto-update from finding and updating a node's ENR to its external socket. 

For security reasons, we can't just accept PONG responses from any node on the network. For this reason, we only ping peers that make it into our routing table as a number of security measures must be passed for a node to enter the table. However if the table is full, we reject peers from entering and therefore do not attempt to PING them. 

This PR suggests the following mechanism:
If a peer fails to make it into the routing table, and we are lacking votes on either IPv4 or IPv6 and this peer can add a vote AND they are an outgoing peer (i.e one that we have chosen to connect to, for security reasons) then we ping them to get an extra vote for our external address.

### DIFF
--- a/src/service.rs
+++ b/src/service.rs
@@ -817,7 +817,7 @@ impl Service {
                         kbucket::Entry::Present(_, status)
                             if status.is_connected() && !status.is_incoming());
 
-                        if should_count {
+                        if should_count | self.require_more_ip_votes(socket.is_ipv6()) {
                             // get the advertised local addresses
                             let (local_ip4_socket, local_ip6_socket) = {
                                 let local_enr = self.local_enr.read();
@@ -1352,19 +1352,10 @@ impl Service {
                         // If we are low on votes and we initiated this connection (i.e it was not
                         // forced on us) then lets get a PONG from this node.
 
-                        if direction == ConnectionDirection::Outgoing {
-                            if let Some(ip_votes) = self.ip_votes.as_mut() {
-                                match (ip_votes.majority(), enr.udp4_socket(), enr.udp6_socket()) {
-                                    // We don't have enough ipv4 votes, but this is an IPv4 node.
-                                    ((Some(_), None), Some(_), _) |
-                                    // We don't have enough ipv6 votes, but this is an IPv6 node
-                                    ((None, Some(_)), _, Some(_)) |
-                                    // We don't have enough ipv6 or ipv4 nodes, ping this peer
-                                    ((None, None), _, _) => self.send_ping(enr, None),
-                                    // We have enough votes do nothing
-                                    ((_, _), _, _) =>  {}
-                                }
-                            }
+                        if direction == ConnectionDirection::Outgoing
+                            && self.require_more_ip_votes(enr.udp6_socket().is_some())
+                        {
+                            self.send_ping(enr, None);
                         }
 
                         self.peers_to_ping.remove(&node_id);
@@ -1554,6 +1545,32 @@ impl Service {
             }
 
             self.connection_updated(node_id, ConnectionStatus::Disconnected);
+        }
+    }
+
+    /// This is a helper function that determines if we need more votes for a specific IP
+    /// class.
+    ///
+    /// If we are in dual-stack made and don't have enough votes for either ipv4 or ipv6 and the
+    /// requesting node/vote is what we need, then this will return true
+    fn require_more_ip_votes(&mut self, is_ipv6: bool) -> bool {
+        if !matches!(self.ip_mode, IpMode::DualStack) {
+            return false;
+        }
+
+        if let Some(ip_votes) = self.ip_votes.as_mut() {
+            match (ip_votes.majority(), is_ipv6) {
+                    // We don't have enough ipv4 votes, but this is an IPv4-only node.
+                    ((None, Some(_)), false) |
+                    // We don't have enough ipv6 votes, but this is an IPv6 node
+                    ((Some(_), None), true) |
+                    // We don't have enough ipv6 or ipv4 nodes, ping this peer
+                    ((None, None), _,) => true,
+                    // We have enough votes do nothing
+                    ((_, _), _,) =>  false,
+                }
+        } else {
+            false
         }
     }
 
--- a/src/service.rs
+++ b/src/service.rs
@@ -1552,7 +1552,7 @@ impl Service {
     /// class.
     ///
     /// If we are in dual-stack mode and don't have enough votes for either ipv4 or ipv6 and the
-    /// requesting node/vote is what we need, then this will return true
+    /// requesting node/vote is what we need, then this will return true.
     fn require_more_ip_votes(&mut self, is_ipv6: bool) -> bool {
         if !matches!(self.ip_mode, IpMode::DualStack) {
             return false;
--- a/src/service.rs
+++ b/src/service.rs
@@ -1566,7 +1566,7 @@ impl Service {
                     ((None, Some(_)), false) |
                     // We don't have enough ipv6 votes, but this is an IPv6 node.
                     ((Some(_), None), true) |
-                    // We don't have enough ipv6 or ipv4 nodes, ping this peer
+                    // We don't have enough ipv6 or ipv4 nodes, ping this peer.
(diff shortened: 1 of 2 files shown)


AgeManning:
cc @pawanjay176 

@ackintosh - What do you think of this logic? Is there an easy way to test this with the current infra?

src/service.rs
```diff
@@ -1345,6 +1345,28 @@ impl Service {
...
+                        // routing table can be full, and so we reject useful IPv6 here.
+                        //
+                        // If we are low on votes and we initiated this connection (i.e it was not
+                        // forced on us) then lets get a PONG from this node.
+
+                        if direction == ConnectionDirection::Outgoing {
+                            if let Some(ip_votes) = self.ip_votes.as_mut() {
+                                match (ip_votes.majority(), enr.udp4_socket(), enr.udp6_socket()) {
+                                    // We don't have enough ipv4 votes, but this is an IPv4 node.
+                                    ((Some(_), None), Some(_), _) |
```
ackintosh, line 1359: Since `ip_votes.majority()` returns `(Option<SocketAddrV4>, Option<SocketAddrV6>)`, should this look like this?

```suggestion
                                    ((None, Some(_)), Some(_), _) |
```

ackintosh:
Currently, we accept a PONG response as a vote only if the node exists our routing table. So the PONG response is ignored even if we PING a node that failed to make it into our routing table. We should also change this condition?

https://github.com/sigp/discv5/blob/765abacf4daae3d41317af3eb9d1f76d826d25be/src/service.rs#L816-L818

ackintosh:
I think Testground doesn’t support IPv6, so we can build a test environment based on Docker Compose. It might require a little work, but I’ll give it a try.

src/service.rs
AgeManning: ahah yeah. I seemed to have confused myself in the last commit. I reversed these. I'll change it back.

AgeManning:
ah nice! Thanks for the catch!

I've modified the logic a bit to also just do this if we are in DualStack mode.

src/service.rs
```diff
@@ -1535,6 +1548,32 @@ impl Service {
...
+    /// class.
+    ///
+    /// If we are in dual-stack made and don't have enough votes for either ipv4 or ipv6 and the
+    /// requesting node/vote is what we need, then this will return true
+    fn require_more_ip_votes(&mut self, is_ipv6: bool) -> bool {
+        if !matches!(self.ip_mode, IpMode::DualStack) {
+            return false;
+        }
+
+        if let Some(ip_votes) = self.ip_votes.as_mut() {
```
jxs, line 1561: nit:
```suggestion
       let Some(ip_votes) = self.ip_votes.as_mut() else {
           return false;
       }
```
instead of returning false after checking the majority and `is_ipv6` conditions, reducing the scope of the if condition of `ip_votes` being `None` and returning early.

src/service.rs
```diff
@@ -1535,6 +1548,32 @@ impl Service {
         }
     }
 
+    /// This is a helper function that determines if we need more votes for a specific IP
+    /// class.
+    ///
+    /// If we are in dual-stack made and don't have enough votes for either ipv4 or ipv6 and the
+    /// requesting node/vote is what we need, then this will return true
```
jxs, line 1555:
```suggestion
    /// requesting node/vote is what we need, then this will return true.
```

src/service.rs
```diff
@@ -1535,6 +1548,32 @@ impl Service {
...
+    fn require_more_ip_votes(&mut self, is_ipv6: bool) -> bool {
+        if !matches!(self.ip_mode, IpMode::DualStack) {
+            return false;
+        }
+
+        if let Some(ip_votes) = self.ip_votes.as_mut() {
+            match (ip_votes.majority(), is_ipv6) {
+                    // We don't have enough ipv4 votes, but this is an IPv4-only node.
```
jxs, line 1563:
```suggestion
                    // We don't have enough ipv4 votes, but this is an IPv4-only node.
```

src/service.rs
```diff
@@ -1535,6 +1548,32 @@ impl Service {
...
+                    ((None, Some(_)), false) |
+                    // We don't have enough ipv6 votes, but this is an IPv6 node
```
jxs, line 1565:
```suggestion
                    // We don't have enough ipv6 votes, but this is an IPv6 node.
```

src/service.rs
```diff
@@ -1535,6 +1548,32 @@ impl Service {
...
+                    ((Some(_), None), true) |
+                    // We don't have enough ipv6 or ipv4 nodes, ping this peer
```
jxs, line 1567:
```suggestion
                    // We don't have enough ipv6 or ipv4 nodes, ping this peer.
```

src/service.rs
```diff
@@ -1535,6 +1548,32 @@ impl Service {
...
+                    ((None, None), _,) => true,
+                    // We have enough votes do nothing
```
jxs, line 1569:
```suggestion
                    // We have enough votes do nothing.
```

src/service/test.rs
```diff
@@ -102,6 +112,65 @@ async fn build_service<P: ProtocolIdentity>(
...
+    enr_key: Arc<RwLock<CombinedKey>>,
+    filters: bool,
+) -> (Service, UnboundedReceiver<HandlerIn>, Sender<HandlerOut>) {
+    let listen_config = ListenConfig::Ipv4 {
+        ip: local_enr.read().ip4().unwrap(),
+        port: local_enr.read().udp4().unwrap(),
+    };
+    let config = ConfigBuilder::new(listen_config).build();
+
+    // Fake's the handler with empty channels
```
jxs, line 126:
```suggestion
    // Fake's the handler with empty channels.
```

src/service/test.rs
```diff
@@ -102,6 +112,65 @@ async fn build_service<P: ProtocolIdentity>(
...
+        local_enr.read().node_id().into(),
+        Duration::from_secs(60),
+        config.incoming_bucket_limit,
+        table_filter,
+        bucket_filter,
+    )));
+
+    let ip_vote = IpVote::new(10, Duration::from_secs(10000));
+
+    // create the required channels
```
jxs, line 149:
```suggestion
    // create the required channels.
```

src/service/test.rs
```diff
@@ -341,3 +410,95 @@ async fn test_handling_concurrent_responses() {
...
+        .build(&enr_key)
+        .unwrap();
+
+    let (mut service, mut handler_recv, _handler_send) = build_non_handler_service(
+        Arc::new(RwLock::new(local_enr)),
+        Arc::new(RwLock::new(enr_key)),
+        false,
+    );
+
+    // Load up the routing table with 100 random ENRs
```
jxs, line 461:
```suggestion
    // Load up the routing table with 100 random ENRs.
```

src/service/test.rs
```diff
@@ -341,3 +410,95 @@ async fn test_handling_concurrent_responses() {
...
+            .ip6(ip)
+            .udp6(DEFAULT_UDP_PORT)
+            .build(&key)
+            .unwrap();
+
+        let direction = ConnectionDirection::Outgoing;
+        service.inject_session_established(enr.clone(), direction);
+    }
+
+    // Collect all the messages to the handler and count the PING requests for ENR v6 addresses
```
jxs, line 490:
```suggestion
    // Collect all the messages to the handler and count the PING requests for ENR v6 addresses.
```

jxs (approved):
LGTM age, left some suggestions

src/service.rs
AgeManning: I think this change is the same. 

If self.ip_votes is None, we immediately return false, we don't check `is_ipv6` or the majority. 

Adding this PR because its probably easier to see this

ackintosh:
I've implemented a simulation (https://github.com/ackintosh/discv5-dual-stack) and tested this PR using it. The ipv6 auto-update works fine!

PR #270, opened 2024-10-15
Increase default vote duration to 2 minutes

AgeManning:
I have been looking into complications with low ipv6 peers in real world networks. 

Originally this default was set, to handle ip changes more rapidly, however with low ipv6 peers on networks this becomes a hindering factor. 

Although it is configurable, the default should be large enough to accommodate real-world networks. 

I think this might have been originally set when we ping'd every connection. As no longer do that, maintaining pong votes for 2 minutes seems to be a reasonable tradeoff.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -161,7 +161,7 @@ impl ConfigBuilder {
     }
 
     /// The interval over which votes are remembered when determining our external IP. A lower
-    /// interval will respond faster to IP changes. Default is 30 seconds.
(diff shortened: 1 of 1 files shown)


jxs (commented):
lower in the file there's still a reference to 30 seconds:
https://github.com/sigp/discv5/blob/9304dbb22ff24d0836d101d6896dc5b565c3a6d8/src/config.rs#L163-L168

AgeManning:
Fixed, thanks man :)

jxs (approved):

PR #271, opened 2024-10-22
Automatic Firewall/NAT Detection

AgeManning:
## Description

Symmetric NATs and IPv6 are currently a pain point in discovery. Discv5 has a mechanism to discover its external IP and port (`socket` from now on). We use a majority voting system based on `PONG` responses. 

The problem is that full-cone NATs can be included in the IP address discovery. Furthermore, with ipv6, all connections are likely not NAT'd and the external IP discovery is going to return our IPv6 address. 

Currently we are assuming that if our external address is discoverable, that we are contactable (as we get discovered when a correct port forward is established in ipv4). However this will generally not be the case for full-cone NATs and IPv6. In fact, if we default to dual-stack, all ipv6 nodes will update their ENR to their ipv6 address, rendering them non-contactable (as ipv6 is the default connection for ipv6-compatible nodes) if their firewall still prevents ipv6 traffic. 

We therefore need a way to determine if we are contactable or not to avoid incorrectly advertising our ipv6 or ipv4 discovered addresses. 

This PR proposes the following:
1. We discover our external IP addresses as usual (ipv6 will likely always be discovered)
2. We advertise our discovered address as usual
3. We wait a period of time for inbound connections
4. If we do not see any inbound connections, we assume we are not contactable and remove our IP advertisement from our ENR and then wait an extended period of time (6 hours) before attempting again
5. If we do see inbound connections, consider ourselves contactable and leave our ENR advertising our address, until we discover a new external ip address (as usual)

I have done some preliminary testing on this PR and it seems to function as I expect. 

However with these added complexity more eyes are always helpful :)

### DIFF
--- a/src/service.rs
+++ b/src/service.rs
@@ -31,6 +31,7 @@ use crate::{
     },
     rpc, Config, Enr, Event, IpMode,
 };
+use connectivity_state::DURATION_UNTIL_NEXT_CONNECTIVITY_ATTEMPT;
 use connectivity_state::{ConnectivityState, TimerFailure};
 use delay_map::HashSetDelay;
 use enr::{CombinedKey, NodeId};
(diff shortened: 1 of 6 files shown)


src/service.rs
```diff
@@ -470,6 +466,26 @@ impl Service {
...
+                        }
+                        TimerFailure::V6 => {
+                            // We have not received enough incoming connections in the required
+                            // time. Remove our ENR advertisement.
+                            info!(ip_version="v6", next_attempt_in=%DURATION_UNTIL_NEXT_CONNECTIVITY_ATTEMPT.as_secs(), "UDP Socket removed from ENR");
+                            if let Err(error) = self.local_enr.write().remove_udp6_socket(&self.enr_key.read()) {
+                                error!(?error, "Failed to update the ENR");
+                            }
+                        }
+                    }
```
ackintosh, line 487: Should we ping_connected_peers to advertise the updated ENR?

ackintosh (commented):
I also tested this, and it works fine!

src/service.rs
AgeManning: Yep, nice catch!

PR #272, opened 2024-10-27
[feat] Made Enr generic with EnrKey in impl

### LINKED ISSUE #233: Make `Enr` generic with `Enr<CombinedKey>` as default
To not limit the flexibility of which key type to use offered by the `Enr` crate, `Enr` should be made generic in discv5. The default generic should still be `CombinedKey`, the key type being used now. This change also requires change the `impl` bodies to be extend to work for all types of keys, not just be implemented for the default `CombinedKey`. This shouldn't be hard because the `EnrKey` trait is at disposal from the enr crate.

```rust
impl<P: ProtocolIdentity, K: EnrKey> Discv5<P, K> { ...
```

mahmudsudo:
## Description

Made Enr generic with EnrKey in impl
## Change checklist

- [x] impl generic
- [x] function "new" args

closes #233

### DIFF
--- a/src/discv5.rs
+++ b/src/discv5.rs
@@ -23,7 +23,7 @@ use crate::{
     service::{QueryKind, Service, ServiceRequest, TalkRequest},
     Config, DefaultProtocolId, Enr, IpMode,
 };
-use enr::{CombinedKey, EnrKey, Error as EnrError, NodeId};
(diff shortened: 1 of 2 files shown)


AgeManning:
Some of the checks are failing, just formatting it looks like at the moment.  Also I think we still want `CombinedKey` to be the default.

AgeManning:
Still seems to be some formatting issues

PR #275, opened 2025-01-15
Prevent dual-stack spamming

AgeManning:
## Description

I've been testing Ipv6 and dual stack support. 

There was a bug, where in dual-stack mode, if a router has not SNAT'd correctly they use random source ports. The `majority()` voting system will never reach a majority and we will continue to ping all peers we see. 

We waste a lot of traffic as we are never going to reach a majority in this scenario. The correct logic is not to check for a majority on a specific port before asking more more PING's rather, if we have enough PONG responses to reach a majority if our NAT was correctly set up, then no longer send more PINGs. 

This PR updates to this new logic.

### DIFF
--- a/src/service/ip_vote.rs
+++ b/src/service/ip_vote.rs
@@ -62,8 +62,8 @@ impl IpVote {
     /// Filter the stale votes and return the majority `SocketAddr` if it exists.
     /// If there are not enough votes to meet the threshold this returns None.
     fn filter_stale_find_most_frequent<K: Copy + Eq + Hash>(
-        &self,
         votes: &HashMap<NodeId, (K, Instant)>,
+        minimum_threshold: usize,
     ) -> (HashMap<NodeId, (K, Instant)>, Option<K>) {
         let mut updated = HashMap::default();
         let mut counter: FnvHashMap<K, usize> = FnvHashMap::default();
@@ -80,7 +80,7 @@ impl IpVote {
             let count = counter.entry(*vote).or_default();
             *count += 1;
             let current_max = max.map(|(_v, m)| m).unwrap_or_default();
-            if *count >= current_max && *count >= self.minimum_threshold {
+            if *count >= current_max && *count >= minimum_threshold {
                 max = Some((*vote, *count));
             }
         }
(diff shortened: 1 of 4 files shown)


src/service/ip_vote.rs
```diff
@@ -23,34 +25,61 @@ impl IpVote {
...
+            SocketAddr::V6(socket) => {
+                self.ipv6_votes
+                    .insert(key, (socket, Instant::now() + self.vote_duration));
+            }
+        }
+    }
+
+    /// Returns true if we have more than the minimum number of non-expired votes for a given ip
+    /// version.
+    pub fn less_than_minimum(&mut self) -> (bool, bool) {
```
ackintosh, line 50: I find this method name slightly misleading, since it actually checks whether the number of votes meets the minimum threshold. How about renaming it to `meets_minimum_threshold`?

src/service/ip_vote.rs
```diff
@@ -23,34 +25,61 @@ impl IpVote {
...
-        let mut ip6_count: FnvHashMap<SocketAddrV6, usize> = FnvHashMap::default();
-        for (socket, _) in self.votes.values() {
-            // NOTE: here we depend on addresses being already cleaned up. No mapped or compat
-            // addresses should be present. This is done in the codec.
-            match socket {
-                SocketAddr::V4(socket) => *ip4_count.entry(*socket).or_insert_with(|| 0) += 1,
-                SocketAddr::V6(socket) => *ip6_count.entry(*socket).or_insert_with(|| 0) += 1,
-            }
-        }
+        self.ipv4_votes.retain(|_, v| v.1 > instant);
```
jxs, line 65: does the order import for majority calculation?
I.e why do we have a test function for `test_three_way_vote_draw`. 
We can save 2 iterations if we combine majority calculation and stale votes removal did at 
https://github.com/sigp/discv5/commit/ac644c8ca01f93642749908a78f93d4d52d9c644 but `test_three_way_vote_draw` fails

jxs (commented):
Thanks age left come somments

src/service/ip_vote.rs
AgeManning: Yeah I think your changes are better. Can't remember about that test. I think we wanted to test that if there is a draw it at least returns Some and not None. It probably should be random from the tie. 

Can you make a PR to adjust the logic to what you have here. 

Why don't we just mutate the hashmap in the function rather than returning a new hashmap and re-setting it?

src/service/ip_vote.rs
AgeManning: Yep, fair. Modified the name. Thanks

src/service/ip_vote.rs
jxs: yeah makes sense, updated and added the function to your PR!

jxs (approved):
Going to approve so that if you think it's ready to merge can merge it age!

PR #276, opened 2025-01-17
Fix encode TalkRequest regression during library transition

KolbyML:
## Description

Well updating from the old rlp library to the new one alloy_rlp there was a regression for encoding/decoding TalkRequests, which is required for the Portal Network.

I noticed something was broken when trying to upgrade Trin to alloy back in Apr 5, 2024. I delayed upgrading https://github.com/ethereum/trin/pull/1231 because I didn't have time to debug this issue at the time.

So here is the fix to the regression and a few tests to prevent it from happening again 

## Change checklist

- [x] Self-review
- [x] Tests if relevant

### DIFF
--- a/src/kbucket/bucket.rs
+++ b/src/kbucket/bucket.rs
@@ -410,9 +410,7 @@ where
             // Adjust `first_connected_pos` accordingly.
             match old_status.state {
                 ConnectionState::Connected => {
-                    if self.first_connected_pos.map_or(false, |p| p == pos.0)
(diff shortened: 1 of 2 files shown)


KolbyML:
@jxs @AgeManning ping for review

ackintosh (approved):

AgeManning:
Yeah. This change cooked us also in a variety of ways, but it was outside the library. I thought I had caught them all, but we don't use the `TalkRequest` RPC, so apologies for missing this one.

PR #279, opened 2025-02-12
perf: use NodeContact instead of Enr for sending talk_req()

KolbyML:
## Description

```rust
    pub fn try_from_enr(enr: Enr, ip_mode: IpMode) -> Result<Self, NonContactable> {
        let socket_addr = match ip_mode.get_contactable_addr(&enr) {
            Some(socket_addr) => socket_addr,
            None => return Err(NonContactable { enr }),
        };

        Ok(NodeContact {
            public_key: enr.public_key(),
            socket_addr,
            enr: Some(enr),
        })
    }
```

Every time we were sending a talk request we were converting the `Enr` to `NodeContact`. In doing this operation we call `public_key()` or in other terms we are recovering the public key every time we send a talk request, in Portal we send all our messages over TalkRequests, currently `public_key()` is taking over 8.7% of the samples I get from running my benchmark https://github.com/ethereum/trin/pull/1660, by reusing the NodeContact we can avoid repetitively recovering the public key every time

In the benchmark I am sending 6.7GB of block history, so 8.7 percent is fairly significant

image

## Notes & open questions

@jxs @AgeManning @ackintosh ping for review

### DIFF
--- a/src/discv5.rs
+++ b/src/discv5.rs
@@ -267,6 +267,10 @@ impl<P: ProtocolIdentity> Discv5<P> {
         nodes_to_send
     }
 
+    pub fn ip_mode(&self) -> IpMode {
+        self.ip_mode
+    }
+
     /// Mark a node in the routing table as `Disconnected`.
     ///
     /// A `Disconnected` node will be present in the routing table and will be only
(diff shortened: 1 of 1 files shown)


AgeManning (approved):
Yep. Typically we know an ENR rather than a `NodeContact`, which I imagine is why this was the API. 

The `TalkReq ` feature as far as I know is mainly used by you guys, so feel free to change the public API. 

We will have to make this change with a new version however as it changes the public facing API.

PR #281, opened 2025-04-02
Report expired sessions

AgeManning:
## Description

Users cannot keep track of active sessions as they expire internally in our cache. This PR adds an extra message (breaking the API), that reports when a session has been removed from our internal cache. 

This does change the expiry logic also. Previously we were removing elements on most operations, i.e when we send/receive a message. We now only remove old elements when a new session is added, or one fails. So long as there are new connections, we will expire old ones. If the peer set remains consistent without errors, we will no longer expire elements simply based on time. I don't consider this an issue (we can also add this functionality back if anyone is concerned about checking the cache more regularly).

### DIFF
--- a/src/lru_time_cache.rs
+++ b/src/lru_time_cache.rs
@@ -37,17 +37,14 @@ impl<K: Clone + Eq + Hash, V> LruTimeCache<K, V> {
     }
 
     /// Retrieves a reference to the value stored under `key`, or `None` if the key doesn't exist.
-    /// Also removes expired elements and updates the time.
     #[allow(dead_code)]
     pub fn get(&mut self, key: &K) -> Option<&V> {
         self.get_mut(key).map(|value| &*value)
     }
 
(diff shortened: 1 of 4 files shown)


src/lru_time_cache.rs
```diff
@@ -37,17 +37,14 @@ impl<K: Clone + Eq + Hash, V> LruTimeCache<K, V> {
     }
 
     /// Retrieves a reference to the value stored under `key`, or `None` if the key doesn't exist.
-    /// Also removes expired elements and updates the time.
     #[allow(dead_code)]
```
KolbyML, line 40:
```suggestion
```

Isn't `#[allow(dead_code)]` not needed as the function is public and hence isn't considered dead by clippy?

KolbyML (approved):
:shipit: looks good that you for resolving this for us so quick :fire:

ackintosh:
Looks good to me.

PR #296, opened 2026-02-24
fix: validate request ID length in Message::decode()

jsvisa:
## Description

`Message::decode()` was constructing RequestId directly from decoded bytes without validating the length. [The discv5 spec limits request IDs to a maximum of 8 bytes](https://github.com/ethereum/devp2p/blob/master/discv5/discv5-wire.md#protocol-messages): 

> The first element of every message-data list is the request ID. request-id is an RLP byte array of length <= 8 bytes. For requests, this value is assigned by the requester. The recipient of a message must mirror the value in the request-id element of the response. The selection of appropriate values for request IDs is left to the implementation.

This caused the node to accept and echo back oversized request IDs (e.g. 9 bytes) in PONG responses, which violates the spec. The node should silently drop messages with invalid request IDs.

Fixes the hive devp2p PingLargeRequestID conformance test.

## Change checklist

- [ ] Self-review
- [x] Tests if relevant

### DIFF
--- a/src/rpc.rs
+++ b/src/rpc.rs
@@ -357,7 +357,7 @@ impl Message {
         }
 
         let id_bytes = Bytes::decode(payload)?;
-        let id = RequestId(id_bytes.to_vec());
+        let id = RequestId::decode(id_bytes.to_vec())?;
 
         let message = match msg_type {
             1 => {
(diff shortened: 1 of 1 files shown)


jxs (approved):
Hi, and thanks for this! LGTM

ackintosh (approved):
LGTM!

PR #300, opened 2026-04-26
Fix: ipv4/ipv6_contactable metrics never updated

ackintosh:
## Description

- Fix `METRICS.ipv4_contactable` and `METRICS.ipv6_contactable` never being updated due to an inverted condition in `Config::build()`
- Add regression tests for both metrics
- ~~Fix `cargo clippy` errors~~
  - ~~Suppress `clippy::collapsible_match` with explanatory comments where PendingEntry::value()` requires `&mut self` and cannot be used in a pattern guard~~
  - (Edit) The clippy fixes have been split into https://github.com/sigp/discv5/pull/301

#### Background  

`Config::build()` contained an inverted boolean condition:  

```rust
// Intended: disable auto-NAT when enr_update is false
if self.config.enr_update {   // ← should be `if !self.config.enr_update`
    self.config.auto_nat_listen_duration = None;
}
```

Because `enr_update` defaults to true, this condition was always true, so `auto_nat_listen_duration` was unconditionally overwritten with `None`. As a result:
- `ConnectivityState` never started an incoming-connection timer
- `received_incoming_connection()` always [returned early](https://github.com/sigp/discv5/blob/6ef4928ed5f9c53df2d95f38a84769dfabf1b617/src/service/connectivity_state.rs#L122-L125)
- `METRICS.ipv4_contactable` and `METRICS.ipv6_contactable` remained false permanently regardless of actual reachability   

## Change checklist

- [x] Self-review
- [x] Tests if relevant

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -352,7 +352,7 @@ impl ConfigBuilder {
 
         // If enr-update is set to false, then it is non-intuitive for discv5 to revoke ENR details
         // when determining NAT status. So we will not do this.
-        if self.config.enr_update {
+        if !self.config.enr_update {
             self.config.auto_nat_listen_duration = None;
         }
 
--- a/src/discv5.rs
+++ b/src/discv5.rs
@@ -20,6 +20,7 @@ use crate::{
     },
     node_info::{NodeAddress, NodeContact},
     service::{QueryKind, Service, ServiceRequest, TalkRequest},
+    socket::UnrecognizedFrame,
     Config, Enr, IpMode,
 };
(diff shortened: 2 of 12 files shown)


jxs (changes requested):
Hi, Akihito, thanks for looking into this! Can we separate the clippy PR into a separate PR? I am also wondering if we can test in a more isolated way as this test is quite complex and needs flags to not affect other tests

jxs:
can you solve the merge conflicts @ackintosh? Should be good to go after

ackintosh:
@jxs I've updated this PR, thanks!

jxs (approved):
LGTM

PR #301, opened 2026-04-27
Fix clippy errors

ackintosh:
## Description

This PR fixes the following clippy errors:
- https://rust-lang.github.io/rust-clippy/rust-1.95.0/index.html#iter_kv_map
- https://rust-lang.github.io/rust-clippy/rust-1.95.0/index.html#collapsible_match

It also removes unnecessary `mut` usages in:
- 3a6f723f37435e05c1f7cce2f818ae910048408e
- b13cb7c793bbfc0db2a6606a5224fdc10688d719 

Note: in b13cb7c793bbfc0db2a6606a5224fdc10688d719,  changed `PendingNode::value()` and `PendingEntry::value()` were updated to avoid requiring mutable access unnecessarily. Previously, the following code caused a compilation error at service.rs:623:

```rust
kbucket::Entry::Pending(ref entry, _) if entry.value().seq() < enr_seq => {
//                                       ^^^^^ cannot borrow immutable local variable `entry` as mutable
    let enr = entry.value().clone();
    to_request_enr = Some(enr);
}
```

## Change checklist

- [x] Self-review
- [x] Tests if relevant

### DIFF
--- a/src/kbucket/bucket.rs
+++ b/src/kbucket/bucket.rs
@@ -90,8 +90,8 @@ impl<TNodeId, TVal: Eq> PendingNode<TNodeId, TVal> {
         self.node.status
     }
 
-    pub fn value_mut(&mut self) -> &mut TVal {
-        &mut self.node.value
+    pub fn value(&self) -> &TVal {
+        &self.node.value
     }
 
     pub fn set_ready_at(&mut self, t: Instant) {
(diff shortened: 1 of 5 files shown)


jxs (approved):
LGTM Akihito, we are touching public API's which I feel don't need to be public, let's address that on a future PR before releasing a new minor version

PR #302, opened 2026-04-28
fix: restore `&mut` APIs for `PendingEntry`

klkvr:
## Description

https://github.com/sigp/discv5/pull/301 removed APIs exposing mutable access to value of `PendingEntry`

## Change checklist

- [ ] Self-review
- [ ] Tests if relevant


klkvr:
cc @jxs would appreciate if this could make it into the next release 🙏

jxs (approved):
ah right sorry! 

> cc @jxs would appreciate if this could make it into the next release 🙏

yeah, I just want us to clean the API and un `pub` unrequired methods, this one was in that train as I thought it was not used.

PR #304, opened 2026-05-15
Reduce public api surface

### LINKED ISSUE #303: Reduce public API surface
## Background

As part of ongoing maintenance, we are reviewing the crate's public API surface to identify items that were exposed unintentionally or as a side effect of the module structure, rather than by deliberate design.

Reducing the public API has several benefits:

- Lowers the semver compatibility burden — internal restructuring no longer requires a breaking version bump
- Makes the intended entry points clearer for new users
- Reduces the risk of users depending on implementation details that may change

We'd like to hear from users **before** making any changes. If you rely on any of the items listed below, please comment with your use case.

---

## Candidates for making private (`pub(crate)`)

The items below are currently `pub` but appear to have no purpose outside the crate itself. We are considering making them `pub(crate)`.

### 1. `pub mod handler` — entire module

The `handler` module manages low-level session state, encryption, and the WHOAREYOU handshake. None of its types are part of the intended user API.

Currently exposed:
- `HandlerIn`, `HandlerOut` — internal message-passing enums between tasks
- `Handler` — the session management task itself
- `WhoAreYouRef`, `Challenge` — internal handshake data structures

**Note:** `NodeAddress` and `NodeContact` are re-exported from `handler` and currently appear in the public API (`Event::SessionsExpired`, `Discv5::talk_req`). If this module is made private, these two types would be re-exported directly from the crate root.

**Question for users:** Do you import anything from `discv5::handler` directly?

---

### 2. `pub mod rpc` — entire module

The `rpc` module contains the wire-format message types for the Discovery v5 protocol. Ordinary users of the crate never need to construct or decode RPC messages manually.

Currently exposed:
- `Message`, `Request`, `Response` — RLP wire-format types
- `RequestBody`, `ResponseBody` — message body enums (Ping/FindNode/Talk variants)
- `RequestId` — appears in `TalkRequest::id()`, so this type specifically would need to remain `pub` or be re-exported

**Question for users:** Do you use `discv5::rpc::*` types directly? If so, for what purpose?

---

### 3. `pub mod packet` — most types

The `packet` module contains the raw UDP packet encoding layer. `ProtocolIdentity` is already re-exported from the crate root and is the only type users are expected to configure.

Currently exposed but likely unnecessary:
- `Packet`, `PacketHeader` — raw packet structures
- `ChallengeData` — internal handshake authenticated data
- `IV_LENGTH`, `STATIC_HEADER_LENGTH`, `MESSAGE_NONCE_LENGTH`, `ID_NONCE_LENGTH` — protocol constants

**Question for users:** Do you use any `discv5::packet::*` types or constants directly?

---

### 4. `service::ServiceRequest`, `service::Service`, `service::CallbackResponse`, `service::QueryKind`

The `service` module is `pub`, which leaks internal implementation types:

- `ServiceRequest` — the internal enum used to send commands from `Discv5` to the background `Service` task
- `Service` — the background service task struct itself
- `CallbackResponse`, `QueryKind` — internal response/query bookkeeping

`TalkRequest` and `Pong` (the return type of `Discv5::send_ping`) are legitimately public and would remain so.

**Question for users:** Do you reference `discv5::service::ServiceRequest` or `discv5::service::Service` directly?

---

### 5. `socket::Socket`, `socket::SocketConfig`, `socket::RecvHandlerConfig`, `socket::InboundPacket`, `socket::RecvPacket`, `socket::OutboundPacket`

The `socket` module is `pub`, but most of its types are internal socket management infrastructure. The intended public surface from this module is already re-exported at the crate root: `ListenConfig`, `RateLimiter`, `RateLimiterBuilder`.

- `Socket` — internal UDP socket wrapper
- `SocketConfig`, `RecvHandlerConfig` — internal configuration for socket tasks
- `InboundPacket`, `RecvPacket`, `OutboundPacket` — internal packet passing types

`UnrecognizedFrame` must remain `pub` as it appears in `Event::UnrecognizedFrame`.

**Question for users:** Do you use any of the above `discv5::socket::*` types directly?

---

### 6. `Discv5::raw_metrics()` and `metrics::InternalMetrics`

`Discv5::raw_metrics()` returns a `&'static InternalMetrics`, which exposes the live atomic counters used internally. The intended public interface for metrics is `Discv5::metrics()`, which returns a clean `Metrics` snapshot struct.

`InternalMetrics` exposes `AtomicUsize` / `AtomicBool` fields and mutation methods (`add_recv_bytes`, `add_sent_bytes`) that are not meaningful from outside the crate.

**Question for users:** Do you call `Discv5::raw_metrics()` or access `metrics::InternalMetrics` directly? If so, what do you need from it that `Discv5::metrics()` does not provide?

---

### 7. `kbucket` module — `IpTableFilter` / `IpBucketFilter`

The mutation side of `KBucketsTable` (`insert_or_update`, `update_node`, `update_node_status`, `entry()` with `PresentEntry`/`AbsentEntry`) and `KBucket` access via `buckets_iter()` / `get_bucket()` are **not candidates for privatization**: [reth](https://github.com/paradigmxyz/reth) uses `KBucketsTable` directly to implement its discv4 routing table (`crates/net/discv4`), relying on the full mutation API and `KBucket::num_entries` / `num_connected` for statistics.

The remaining candidates are:

- `IpTableFilter`, `IpBucketFilter` — `pub` but the `ip_limit` config option applies them automatically; reth does not use either
- `kbucket::Filter` trait — no known external implementation

**Question for users:** Do you implement the `kbucket::Filter` trait for custom filtering, or use `IpTableFilter` / `IpBucketFilter` directly?

---

## Items we do NOT plan to change

The following items are intentionally public and will remain so regardless of the outcome of this discussion:

| Item | Reason |
|------|--------|
| `Discv5` and all its methods | Primary user API |
| `Event` | Event stream API |
| `Config`, `ConfigBuilder` | Configuration API |
| `Error`, `RequestError`, `QueryError`, `ResponseError` | Error handling |
| `Executor`, `TokioExecutor` | Custom runtime support |
| `IpMode` | Exposed via `Discv5::ip_mode()` and `Config` |
| `PermitBanList` | Used in `Config` and `Discv5::ban_*` / `permit_*` methods |
| `TalkRequest` | Received via event stream |
| `ListenConfig`, `RateLimiter`, `RateLimiterBuilder` | Configuration API |
| `ProtocolIdentity` | Used in `Config` |
| `Metrics` | Returned by `Discv5::metrics()` |
| `Key`, `ConnectionState`, `ConnectionDirection` | Used in `Discv5::table_entries()` and routing table inspection |
| `NodeAddress`, `NodeContact` | Used in `Event::SessionsExpired` and `Discv5::talk_req()` |
| `Pong` | Returned by `Discv5::send_ping()` |
| `UnrecognizedFrame` | Used in `Event::UnrecognizedFrame` |
| `KBucketsTable` (read API) | Returned by `Discv5::kbuckets()` |
| `RequestId` | Returned by `TalkRequest::id()` |

---

## How to respond

Please comment if any of the following applies to you:

1. You import a type listed as a candidate and would be broken by making it private
2. You have a legitimate use case for a candidate that we haven't considered
3. You believe an item in the "not planning to change" list should also be reconsidered

We aim to make a decision within a few weeks after gathering feedback.

ackintosh:
## Description

closes https://github.com/sigp/discv5/issues/303

## Change checklist

- [x] Self-review
- [x] Tests if relevant


src/socket/mod.rs
```diff
@@ -23,8 +23,9 @@ pub use filter::{
     rate_limiter::{RateLimiter, RateLimiterBuilder},
     FilterConfig,
 };
-pub use recv::{InboundPacket, RecvPacket, UnrecognizedFrame};
-pub use send::OutboundPacket;
+pub use recv::UnrecognizedFrame;
+pub(crate) use recv::{InboundPacket, RecvPacket};
+pub(crate) use send::OutboundPacket;
```
jxs: do we need the pub(crate) use? I'd skip this import redirection

src/kbucket.rs
```diff
@@ -81,7 +81,7 @@ pub use bucket::{
     ConnectionState, FailureReason, InsertResult as BucketInsertResult, UpdateResult,
     MAX_NODES_PER_BUCKET,
 };
-pub use filter::{Filter, IpBucketFilter, IpTableFilter};
+pub(crate) use filter::{Filter, IpBucketFilter, IpTableFilter};
```
jxs: same as above, do we need the `pub use` here?

jxs (commented):
thanks for this Akihito,
Overall looks good to me! Left just two comments

CC @klkvr @KolbyML @onbjerg @mattsse @kevaundray @theochap @morph-dev @jmezo @njgheorghita @DaniPopes @Vid201 @carver @mskrzypkows @jsvisa @Nesopie @Padraic-O-Mhuiris @ogenev @pedrohba1 @mahmudsudo @0xcrust @LNSD @arsenron @e3243eric @xiaoxianBoy

last call before this gets merged

src/socket/mod.rs
ackintosh: These are used within the `handler` module. To remove the `pub(crate) use`, we'd need to expose the modules themselves at the crate level instead:

```diff
- mod recv;
+ pub(crate) mod recv;
- mod send;
+ pub(crate) mod send;
```

jxs: yeah that's what I was suggesting

src/kbucket.rs
ackintosh: The filter structs are used in the crate root (`discv5.rs`), so I think we either need to keep the `pub(crate) use` or make the `filter` module itself `pub(crate)`.

jxs: yeah let's make the `filter` module `pub(crate)` wdyt?

ackintosh:
Updated this PR, thanks for your feedback @jxs!

I also added https://github.com/sigp/discv5/pull/304/changes/8d04d7f52fd66a146f4b0d2b1fc3ff52dd332025 as a small usability improvement.

jxs (approved):
Thanks Akihito!

PR #309, opened 2026-07-30
fix: resend active WHOAREYOU challenge

mattsse:
Resends the currently active WHOAREYOU packet when another ordinary packet arrives from the same node and endpoint during an unfinished handshake.

This preserves the original request nonce and challenge data instead of dropping the retry or creating another outstanding challenge. It also adds a regression test for the geth/Hive `HandshakeResend` flow.


ackintosh (approved):
Looks good to me. Thank you @mattsse 🙏

ackintosh:
The clippy errors will be fixed by https://github.com/sigp/discv5/pull/310

ackintosh:
Note that the current spec says to respond with a *new* WHOAREYOU challenge.

go-ethereum has re-sent the existing challenge since https://github.com/ethereum/go-ethereum/pull/31356, and I have filed https://github.com/ethereum/devp2p/pull/278 to update the spec accordingly.

src/handler/tests.rs
```diff
@@ -178,6 +178,108 @@ async fn simple_session_message() {
...
+        .unwrap()
+        .unwrap();
+        let response = &response[..response_len];
+        let (challenge, challenge_data) =
+            Packet::decode(&sender_enr.node_id(), ProtocolIdentity::default(), response).unwrap();
+        assert!(matches!(
+            challenge.header.kind,
+            PacketKind::WhoAreYou { .. }
+        ));
+        assert_eq!(challenge.header.message_nonce, first_nonce);
```
jxs, line 269: since we're here we might as well confirm the id_nonce in the WHOAREYOU header matches

jxs (commented):
LGTM, some minor comment

ackintosh:
@mattsse  Could you take a quick look at @jxs’s minor comment? 🙏  Also, the corresponding [spec PR](https://github.com/ethereum/devp2p/pull/278) has been merged now, so we should be aligned there too.

src/handler/tests.rs
Oluwajakul: Hi

jxs (approved):
LGTM thanks!

